PR #1, opened 2017-08-09
Setup Travis CI

BertrandF23:


.travis.yml
```diff
@@ -0,0 +1,28 @@
...
+matrix:
+  allow_failures:
+    - php: 7.1
+
+before_install:
+  - echo "extension = mongodb.so" >> ~/.phpenv/versions/$(phpenv version-name)/etc/php.ini
+
+before_script:
+  - composer self-update
+  - composer install --prefer-source --no-interaction --dev
```
campru: In there some reason to use `--prefer-source` instead of `--prefer-dist`? And also `--dev` if it is not added is not a problem because is the default behaviour

.travis.yml
```diff
@@ -0,0 +1,28 @@
...
+
+script:
+  - bin/phpunit --coverage-text --debug
+
+notifications:
+  email:
```
campru: To be sure, I would suggest to add the email "si-integration@smartandco.com"

BertrandF23: How about: ?

notifications:
  email:
    recipients:
      - si-integration@smartandco.com
    on_success: change 
    on_failure: always

It will  send an email in case of success but only after the 1st time it passes (since the last failure)

campru: It is fine for me, good way to avoid to send emails for sucessfull builds

PR #4, opened 2017-08-22
Split json file loader response by type. 

melmccann:
Updated and ready for review.
Thanks,
Mel


Components/JsonFileLoader/JsonFileLoaderProducer.php
```diff
@@ -18,6 +18,9 @@ class JsonFileLoaderProducer extends Producer
 {
     const OPTION_FILENAME = 'filename';
     const OPTION_BASE_PATH = 'base_path';
+    const OPTION_TYPE = 'type';
```
Thevart: We don't need to have this constant on the producer as they are defined in the protocol.

Thevart: We should remove all the constraint defined in this producer

PR #8, opened 2017-08-28
Added mapping method to remove new lines

BertrandF23:


Tools/Mapper/Mapper.php
```diff
@@ -339,5 +339,16 @@ public function emptyTo($string, $value)
...
+
+    /**
+     * Remove any new lines in $string, like \n, \n\r or \r
+     * @param $string
+     * @return mixed
+     */
+    public function removeNewLines($string)
+    {
+        $string = preg_replace('/[\r\n]+/', ' ', trim($string));
+        return $string;
```
campru: This can be simplified as return preg_replace('/[\r\n]+/', ' ', trim($string));

PR #13, opened 2017-10-09
Add db_connection_name in dbal protocol

campru:
- Use db_connection_name value in DbalStepsProvider
- Add UT for DbalStepsProvicer and ConfigurableDbalProtocol
- Fix namespace in RestConfigurableProducerTest
- Run php-cs-fixer to the php files modified

### DIFF
--- a/Components/DB/Dbal/DbalStepsProvider.php
+++ b/Components/DB/Dbal/DbalStepsProvider.php
@@ -97,7 +97,7 @@ class DbalStepsProvider extends Service implements ConfigurableStepsProviderInte
      */
     public function executeStep($stepAction, array &$stepActionParams, array &$options, array &$context)
     {
-        if ($this->doctrine === null) {
+        if (null === $this->doctrine) {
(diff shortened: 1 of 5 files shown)


Components/DB/Dbal/DbalStepsProvider.php
```diff
@@ -278,10 +278,17 @@ protected function performQuery(array $configuration, &$context, array $paramete
...
      * @param $context
+     *
      * @return object
      */
     protected function getConnection($context)
     {
-        return $this->doctrine->getConnection();
+        $connectionName = null;
+
+        if (!empty($context['options']['db_connection_name'])) {
```
BertrandF23: I am not sure that with that name we are talking about a connection name or a database name...
Because in the Platform Bundle, when we use 
$databaseName = $context['options']['db_connection_name'];
it means that db_connection_name is a database name.

BertrandF23: I am afraid there will be confusion between database and connection, and the way they are setup in databases.yml

PR #15, opened 2017-10-16
Fix tests or skip them for RabbitMq

Thevart:


.travis.yml
```diff
@@ -21,7 +21,8 @@ before_install:
   - echo "extension = mongodb.so" >> ~/.phpenv/versions/$(phpenv version-name)/etc/php.ini
 
 before_script:
-  - if [[ "$(php -v | grep 'PHP 7')" ]]; then yes | pecl install -f apcu; fi
+  - bash travis/install-apcu-pecl.sh || true
```
campru: `|| true` should be removed

.travis.yml
```diff
@@ -21,7 +21,8 @@ before_install:
...
+  - phpenv config-add travis/travis.php.ini
```
campru: I think this step can be moved in before_install part instead of before_script. The filename should be called `travis/php.ini`

PR #22, opened 2017-11-09
Close database connections after execute.

melmccann:
This prevents the MySql has gone away error but is a temporary fix as it will impact performance. 
Please review.
Mel


BertrandF23:
Closed at the request of Mel.

PR #31, opened 2018-02-02
Add functionality for callbacks

melmccann:
Base functionality for allowing callbacks in async flows


Core/Handlers/MessageHandler.php
```diff
@@ -29,6 +29,7 @@
 use Smartbox\Integration\FrameworkBundle\DependencyInjection\Traits\UsesEndpointFactory;
 use Symfony\Component\DependencyInjection\ContainerAwareInterface;
 use Symfony\Component\DependencyInjection\ContainerAwareTrait;
+use  Smartbox\Integration\FrameworkBundle\Core\Messages\CallbackExchangeEnvelope;
```
campru: Here you can an extra space. I suggest to pass the php cs fixer

Core/Handlers/MessageHandler.php
```diff
@@ -381,9 +393,15 @@ public function onHandleException(
                 $this->deferExchangeMessage($retryExchangeEnvelope, $this->retryURI);
             } // If it's an exchange that is failing and it should not be retried later
             else {
```
campru: This condition instead of an else it should be an else if and in this case I think it could be simplified like this
```
elseif ($context->get(Context:: CONTEXT_CALLBACK) === true &&
$context->get(Context::CONTEXT_CALLBACK_METHOD) != '')
```
To do this it requires to add in the code a line 
```
$context = $exchangeBackup->getIn()->getContext();
```
I suggest to add this line of code after the dispatch event code and before the first if

Core/Handlers/MessageHandler.php
```diff
@@ -381,9 +393,15 @@ public function onHandleException(
...
+                $headers = $exchangeBackup->getIn()->getContext()->toArray();
+                if($headers && $this->callbackEndpoint && in_array('callback',$headers) && in_array('callbackMethod',$headers) &&  $headers['callback'] == true && $headers['callbackMethod']!='') {
+                    $callbackEnvelope = new CallbackExchangeEnvelope($exchangeBackup, $exception->getProcessingContext());
+                    $this->addCallbackHeadersToEnvelope($callbackEnvelope, $exception, $processor);
+                    $callbackExchange = new Exchange($callbackEnvelope);
+                    $this->callbackEndpoint->produce($callbackExchange);
```
campru: The content inside the if can be simplified using some part of the existing code
```
$callbackExchangeEnvelope = new CallbackExchangeEnvelope($exchangeBackup, $exception->getProcessingContext());
$this->addCallbackHeadersToEnvelope($callbackExchangeEnvelope, $exception, $processor);
this->deferExchangeMessage($callbackExchangeEnvelope, $this->callbackUri);
```

Core/Handlers/MessageHandler.php
```diff
@@ -454,6 +472,28 @@ private function addCommonErrorHeadersToEnvelope(ErrorExchangeEnvelope $envelope
...
+     * @param CallbackExchangeEnvelope $envelope
+     * @param ProcessingException   $exception
+     */
+    private function addCallbackHeadersToEnvelope(CallbackExchangeEnvelope $envelope, ProcessingException $exception, ProcessorInterface $processor)
+    {
+        $originalException = $exception->getOriginalException();
+        $errorDescription = $originalException ? $originalException->getMessage() : $exception->getMessage();
+
+        $envelope->setHeader(CallbackExchangeEnvelope::HEADER_CREATED_AT, round(microtime(true) * 1000));
+        $envelope->setHeader(CallbackExchangeEnvelope::HEADER_ERROR_MESSAGE, $errorDescription);
```
campru: It is missing to add in the envelope the status code of the originalException. In this case needs to be mandatory from the original exception that basically it will be from "UnrecoverableErrorException"

Core/Handlers/MessageHandler.php
melmccann: I did run it, but it didn't remove that :-D

Core/Handlers/MessageHandler.php
```diff
@@ -381,9 +393,15 @@ public function onHandleException(
                 $this->deferExchangeMessage($retryExchangeEnvelope, $this->retryURI);
             } // If it's an exchange that is failing and it should not be retried later
             else {
+                $headers = $exchangeBackup->getIn()->getContext()->toArray();
+                if($headers && $this->callbackEndpoint && in_array('callback',$headers) && in_array('callbackMethod',$headers) &&  $headers['callback'] == true && $headers['callbackMethod']!='') {
+                    $callbackEnvelope = new CallbackExchangeEnvelope($exchangeBackup, $exception->getProcessingContext());
+                    $this->addCallbackHeadersToEnvelope($callbackEnvelope, $exception, $processor);
+                    $callbackExchange = new Exchange($callbackEnvelope);
+                    $this->callbackEndpoint->produce($callbackExchange);
+                }
```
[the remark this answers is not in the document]
melmccann: cool. I will add these changes.

Core/Handlers/MessageHandler.php
melmccann: added

Core/Exchange.php
```diff
@@ -20,6 +20,8 @@ class Exchange implements SerializableInterface
     const HEADER_PARENT_EXCHANGE = 'parent_exchange';
     const HEADER_HANDLER = 'handler';
     const HEADER_FROM = 'from';
+    const HEADER_METHOD = 'method';
+    const HEADER_STATUS = 'status';
```
campru: These properties can be removed from here and be added in platform if there are needed just for the callback success or failure

PR #40, opened 2018-05-03
New 'debug' step, a nasty hack to be able to easily display context v…

BertrandF23:
Example of use inside yaml files:

              - debug:
                  sleep: 10 # How long do you want to wait in seconds?
                  display:
                      - "context['vars']['salesOrder']" # This is what I want to display within the context
                      - "context['body']"
                      - "context['responses']
                  break: true # Will add a breakpoint in a place where I will be able to watch the whole context. Whaooow.
                  exit: true # I don't want to go further!


Configurability/ConfigurableServiceHelper.php
```diff
@@ -224,6 +232,43 @@ public function validate(array $stepConfig, array &$context)
...
+
+            if ('sleep' == $action) {
+                echo 'Sleeping for '.$parameters.' seconds due to sleep step...'."\n";
+                sleep($parameters);
+            }
+
+            if ('exit' == $action && false !== $parameters) {
+                echo 'Exit due to debug step'."\n";
+                exit;
+            }
```
moonraking: could also add an action to do something like this
https://github.com/smartboxgroup/integration-framework-bundle/blob/master/Core/Processors/Processor.php#L140

BertrandF23: Done, great idea.

PR #41, opened 2018-05-09
Add new event for unrecoverable failed exchange

Thevart:


Core/Handlers/MessageHandler.php
```diff
@@ -299,26 +299,36 @@ protected function prepareExchange(Exchange $ex)
         }
     }
 
+
     /**
+     * Dispatch handler event depending
```
moonraking: Is this a complete comment?

PR #44, opened 2018-05-22
Allowing the response exception message to be the original one

andreymor:
Using a producer config parameter on request (display_error), the SOAP and REST APIs should return the message and code from the original exception.

### DIFF
--- a/Components/WebService/Rest/Exceptions/RestException.php
+++ b/Components/WebService/Rest/Exceptions/RestException.php
@@ -91,6 +91,9 @@ class RestException extends \Exception implements SerializableInterface, Externa
         $this->responseHttpHeaders = $responseHeaders;
         $this->responseHttpBody = $responseBody;
         $this->responseStatusCode = $responseStatusCode;
+
+        $this->originalCode = $code;
+        $this->originalMessage = json_decode($this->responseHttpBody, true)['message'];
     }
 
     /**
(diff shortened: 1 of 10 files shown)


Configurability/ConfigurableServiceHelper.php
```diff
@@ -98,13 +96,11 @@ public function createContext(array $options, MessageInterface $message = null,
         return $context;
     }
 
-    public function resolveArray($input, &$context)
-    {
+    public function resolveArray($input, &$context){
```
melmccann: please run `php-cs-fixer fix vendor/smartbox/integration-framework-bundle/Configurability/ConfigurableServiceHelper.php --rules=@Symfony`.
The { should be on the next line.

Components/WebService/Rest/Exceptions/RestException.php
```diff
@@ -91,6 +91,9 @@ public function __construct(
         $this->responseHttpHeaders = $responseHeaders;
         $this->responseHttpBody = $responseBody;
         $this->responseStatusCode = $responseStatusCode;
+
+        $this->originalCode = $code;
+        $this->originalMessage = json_decode($this->responseHttpBody, true)['message'];
```
campru: Do all the responses the json content have a "message" key?

PR #45, opened 2018-05-30
In mapper, added formatDateTimeUtc function to pass through PHP date …

BertrandF23:
…format limitation


campru:
This method shouldn't be needed. In framework there is a place called dictionary that is to define different formats and then we use this format. Basically the change would be to add a new format in that dictionary and start using instead of creating a new method for a specific format. The line where this should be added is the following one. https://github.com/smartboxgroup/integration-framework-bundle/blob/master/Tools/Mapper/Mapper.php#L20

BertrandF23:
As discussed this morning, we need to make sure that the date is in UTC before formatting and adding "Z" at the end. That's why we needed a new method.

PR #49, opened 2018-06-27
Added strtoupper in Mapper

BertrandF23:


melmccann:
Nice job @BertrandF23

PR #53, opened 2018-07-18
Adding SOAP headers as array to be able to be logged as REST headers

andreymor:
To be able to log SOAP headers in the same way as REST headers, now the string is converted to an array

### DIFF
--- a/Components/WebService/Soap/AbstractSoapConfigurableProducer.php
+++ b/Components/WebService/Soap/AbstractSoapConfigurableProducer.php
@@ -296,27 +296,56 @@ abstract class AbstractSoapConfigurableProducer extends AbstractWebServiceProduc
         return '';
     }
 
+    /**
+     * @param $context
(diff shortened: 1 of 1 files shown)


Components/WebService/Soap/AbstractSoapConfigurableProducer.php
```diff
@@ -296,27 +296,56 @@ public static function getConnectivitySmokeTestLabels()
...
+     * @param $headerContent
+     * @return array
+     */
+    private function getHeaderLines($headerContent)
+    {
+        $headerLines = [];
+        $headers = array_filter(preg_split("(\r\n|\r|\n)", $headerContent));
+
+        foreach($headers as $line) {
+            if(false === strpos($line, 'POST')) {
```
BertrandF23: It seems that phpcsfixer is not setup on your system...

PR #61, opened 2018-09-03
Adding new parse to soap headers to make it possible to be logged as …

andreymor:
- Adding a trait to make possible using the parser for all necessary classes;
- Using the new method to parse the headers to array on any soap exception;

### DIFF
--- a/Components/WebService/Soap/ParseHeadersTrait.php
+++ b/Components/WebService/Soap/ParseHeadersTrait.php
@@ -1,0 +1,32 @@
+<?php
+
+namespace Smartbox\Integration\FrameworkBundle\Components\WebService\Soap;
+
+/**
+ * Trait ParseHeadersTrait.
+ */
+trait ParseHeadersTrait
+{
+    /**
+     * @param $headerContent
+     *
+     * @return array
+     */
+    private function parseHeadersToArray($headerContent)
+    {
+        $headerLines = [];
+        $headers = array_filter(preg_split("(\r\n|\r|\n)", $headerContent));
+
+        foreach ($headers as $line) {
+            if (false === strpos($line, 'POST')) {
+                $values = explode(':', $line, 2);
+
+                if (isset($values[1]) && isset($values[0]) && !empty($values[0] && !empty($values[1]))) {
+                    $headerLines[$values[0]] = trim(str_replace('"', '', $values[1]));
+                }
+            }
+        }
+
+        return $headerLines;
+    }
+}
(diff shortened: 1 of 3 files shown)


Components/WebService/Soap/ParseHeadersTrait.php
```diff
@@ -0,0 +1,32 @@
...
+     *
+     * @return array
+     */
+    private function parseHeadersToArray($headerContent)
+    {
+        $headerLines = [];
+        $headers = array_filter(preg_split("(\r\n|\r|\n)", $headerContent));
+
+        foreach ($headers as $line) {
+            if (false === strpos($line, 'POST')) {
```
GaryPEGEOT: This will not work if method if `!==` than POST. I think it would be a better idea to drop the first row of the `$headers` array

Components/WebService/Soap/ParseHeadersTrait.php
```diff
@@ -0,0 +1,32 @@
+<?php
+
+namespace Smartbox\Integration\FrameworkBundle\Components\WebService\Soap;
+
+/**
+ * Trait ParseHeadersTrait.
+ */
+trait ParseHeadersTrait
```
GaryPEGEOT: Can you add a Unit test for this trait?

andreymor: Doneeee

Components/WebService/Soap/ParseHeadersTrait.php
andreymor: Its not only the first line. 
I have added a validation for other HTTP methods

Tests/Unit/Components/WebService/Soap/ParseHeadersTraitTest.php
```diff
@@ -0,0 +1,91 @@
...
+    /**
+     * @group parseSoapHeaders
+     * @dataProvider provideUnparseableData
+     */
+    public function testUnParseHeadersToArray($data)
+    {
+        $result = $this->parseTrait->parseHeadersToArray($data);
+
+        $this->assertTrue(is_array($result), 'The parser should return an array');
+        $this->assertTrue(0 == count($result), 'The parser should return an empty array');
```
campru: There are in PHPUnit asserts to check count and is array. This could be done in this way
```
$this-> assertInternalType('array', $resul);
$this->assertCount(0, $result);
```

Tests/Unit/Components/WebService/Soap/ParseHeadersTraitTest.php
```diff
@@ -0,0 +1,91 @@
...
+    /**
+     * @group parseSoapHeaders
+     * @dataProvider provideParseableData
+     */
+    public function testParseHeadersToArray($header, $expected)
+    {
+        $result = $this->parseTrait->parseHeadersToArray($header);
+
+        $this->assertTrue(is_array($result), 'The parser should return an array');
+        $this->assertTrue(count($result) > 0, 'The parser should not return an empty array');
```
campru: Instead of using assert true and internally functions you could use specific asserts
```
$this->assertInternalType('array', $result);
$this->assertNotEmpty($result);
```

PR #66, opened 2018-09-10
Bugfix/request headers temp fix

melmccann:
Adding a temporary fix to deal with serialization of older versions of events that had requestHeaders as a string.


moonraking:
Please do not merge this.

PR #67, opened 2018-09-12
Display time when starting and stopping a consumer

BertrandF23:


BertrandF23:
Now in log files, this is for instance what we have:
`Consumer was gracefully stopped for: queue://main/api
Consuming from queue://main/api limited to 100 messages.`
The goal is to have timing information:
`2018-09-12 10:30:26.457900 Consumer was gracefully stopped for: queue://main/api
2018-09-12 10:30:35.319800 Consuming from queue://main/api limited to 100 messages.`

Components/FileService/Csv/CsvConfigurableStepsProvider.php
```diff
@@ -347,16 +347,16 @@ protected function readLines(array &$stepActionParams, array &$endpointOptions,
...
+            if (false === $row) {
                 break;
             }
 
             $rows[] = $row;
             ++$i;
         }
 
-        if (count($rows) === 0) {
-            throw new NoResultsException("No more results from $$fullPath");
```
BertrandF23: This is the only real change (silly mistake ;-) )

PR #69, opened 2018-09-17
Deal with the posibility of microtime returing 0 microsecounds

moonraking:


Components/DB/DBConfigurableConsumer.php
```diff
@@ -131,9 +127,13 @@ public function consume(EndpointInterface $endpoint)
                 $endpoint->handle($message);
 
                 if ($this->logger) {
+                    $microTime = number_format(microtime(true), 6, '.', '');
```
andreymor: Why not create a method on Abstract and calls here?
Like consumerLogMessage();
The code is the same as for the message.

moonraking: Good idea

PR #73, opened 2018-10-17
We should catch the exception when wsdl download failed and then retr…

melmccann:
…y to send the message again by sending back to the retry queue.

### DIFF
--- a/Components/WebService/Soap/AbstractSoapConfigurableProducer.php
+++ b/Components/WebService/Soap/AbstractSoapConfigurableProducer.php
@@ -177,18 +177,20 @@ abstract class AbstractSoapConfigurableProducer extends AbstractWebServiceProduc
         $soapHeaders = isset($params[self::SOAP_HEADERS]) ? $params[self::SOAP_HEADERS] : [];
         $httpHeaders = isset($params[self::HTTP_HEADERS]) ? $this->confHelper->resolve($params[self::HTTP_HEADERS], $context) : [];
 
-        $soapOptions['connection_timeout'] = $endpointOptions[ConfigurableWebserviceProtocol::OPTION_CONNECT_TIMEOUT];
-        if ($this->getSoapClient($endpointOptions)) {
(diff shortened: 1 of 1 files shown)


Components/WebService/Soap/AbstractSoapConfigurableProducer.php
```diff
@@ -177,18 +177,20 @@ protected function request(array &$stepActionParams, array &$endpointOptions, ar
...
+            $soapOptions['connection_timeout'] = $endpointOptions[ConfigurableWebserviceProtocol::OPTION_CONNECT_TIMEOUT];
+            if ($this->getSoapClient($endpointOptions)) {
+                $httpHeaders[self::HTTP_HEADER_TRANSACTION_ID] = $context['msg']->getContext()['transaction_id'];
+                $httpHeaders[self::HTTP_HEADER_EAI_TIMESTAMP] = $context['msg']->getContext()['timestamp'];
+                $soapClient = $this->getSoapClient($endpointOptions);
                 $soapClient->setRequestHeaders($httpHeaders);
-            }catch (\SoapFault $exception){
-                $this->throwRecoverableSoapFaultException($exception->getMessage(), Response::HTTP_INTERNAL_SERVER_ERROR, $exception);
             }
+        } catch (\ErrorException $exception) {
```
GaryPEGEOT: I don't like catching `\ErrorException` like that. It mean if someone misuse a function it will be caught here and will be hard to debug.

GaryPEGEOT: I don't like catching `\ErrorException` like that. It mean if someone misuse a function it will be caught here and will be hard to debug.

GaryPEGEOT: I don't like catching `ErrorException` here, it means that if someone misuse a function it will be caught and marked a SoapFault...

GaryPEGEOT: I don't like catching `ErrorException` here, it means that if someone misuse a function it will be caught and marked a SoapFault...

PR #74, opened 2018-11-02
Add rollback function. To be used when exceptions happen and transact…

melmccann:
…ions are left open.

### DIFF
--- a/Components/DB/Dbal/DbalStepsProvider.php
+++ b/Components/DB/Dbal/DbalStepsProvider.php
@@ -291,4 +291,12 @@ class DbalStepsProvider extends Service implements ConfigurableStepsProviderInte
 
         return $this->doctrine->getConnection($connectionName);
     }
+
+    /**
+     * @param $context
+     */
+    protected function rollback(&$context)
+    {
+        $this->getConnection($context)->executeQuery('ROLLBACK', [], []);
+    }
 }


Components/DB/Dbal/DbalStepsProvider.php
```diff
@@ -291,4 +291,12 @@ protected function getConnection($context)
 
         return $this->doctrine->getConnection($connectionName);
     }
+
+    /**
+     * @param $context
+     */
+    protected function rollback(&$context)
+    {
+        $this->getConnection($context)->executeQuery('ROLLBACK', [], []);
```
campru: Doctrine DBAL has a method for the rollback. https://www.doctrine-project.org/projects/doctrine-dbal/en/2.8/reference/transactions.html

Components/DB/Dbal/DbalStepsProvider.php
```diff
@@ -291,4 +291,12 @@ protected function getConnection($context)
 
         return $this->doctrine->getConnection($connectionName);
     }
+
+    /**
+     * @param $context
+     */
+    protected function rollback(&$context)
```
campru: In this case you don't need to pass the context as reference

Components/DB/Dbal/DbalStepsProvider.php
melmccann: yes. I will remove from here. We will handle elsewhere.

Components/DB/Dbal/DbalStepsProvider.php
melmccann: will remove this. Thanks @campru

PR #77, opened 2018-11-21
Change the way we require the serilizer and update core to a dev version

moonraking:


Tests/Functional/Drivers/Queue/StompQueueDriverTest.php
```diff
@@ -86,7 +86,7 @@ public function testShouldSendReceiveAndAckOnce(MessageInterface $msg)
 
         $this->driver->ack();
 
-        sleep(1);
+        \sleep(1);
```
GaryPEGEOT: This should have a special group, see: https://symfony.com/doc/master/components/phpunit_bridge.html#clock-mocking

Tests/Functional/Drivers/Queue/StompQueueDriverTest.php
```diff
@@ -146,7 +146,7 @@ public function testAfterRetriesShouldDiscard($msg)
                 $this->driver->nack();
             }
             $msgOut = $this->driver->receive();
-            sleep(1);
+            \sleep(1);
```
GaryPEGEOT: Same here

Tests/Functional/Drivers/Queue/StompQueueDriverTest.php
```diff
@@ -212,7 +212,7 @@ public function testAfterTtlShouldDiscard(MessageInterface $msg)
         $this->driver->subscribe($this->queueName);
 
         // After 1 second, the message is still there
-        sleep(1);
+        \sleep(1);
```
GaryPEGEOT: Same here: https://symfony.com/doc/master/components/phpunit_bridge.html#clock-mocking

Tests/Functional/StepProviders/CsvConfigurableStepsProviderTest.php
```diff
@@ -45,22 +45,22 @@ public static function setUpBeforeClass()
...
+        \mkdir(self::TMP_FOLDER);
     }
 
     /**
      * {@inheritdoc}
      */
     public static function tearDownAfterClass()
     {
-        $files = glob(self::TMP_FOLDER.'*');
+        $files = \glob(self::TMP_FOLDER.'*');
```
GaryPEGEOT: Even though is not really related to the ticket, using https://symfony.com/doc/current/components/filesystem.html#remove would be nicer here

Tests/Functional/StepProviders/CsvConfigurableStepsProviderTest.php
```diff
@@ -298,13 +298,13 @@ public function testAppendLinesToFile()
 
         //The original file should be there
         $this->assertFileExists(self::TMP_FOLDER.$fileName);
-        $lines = file(self::TMP_FOLDER.$fileName);
-        $this->assertEquals(count($lines), 7); //should be 7 rows
-        $this->assertEquals(count(explode('|', $lines[0])), 3); //should be 3 cols
+        $lines = \file(self::TMP_FOLDER.$fileName);
+        $this->assertEquals(\count($lines), 7); //should be 7 rows
```
GaryPEGEOT: `$this->assertCount` would be a better choice

Tests/Functional/StepProviders/CsvConfigurableStepsProviderTest.php
```diff
@@ -334,12 +334,12 @@ public function testAppendLinesToFileWithHeader()
 
         //The original file should be there
         $this->assertFileExists(self::TMP_FOLDER.$fileName);
-        $lines = file(self::TMP_FOLDER.$fileName);
-        $this->assertEquals(2, count($lines)); //should be 7 rows
-        $this->assertEquals(count(explode('|', $lines[0])), 3); //should be 3 cols
+        $lines = \file(self::TMP_FOLDER.$fileName);
+        $this->assertEquals(2, \count($lines)); //should be 7 rows
```
GaryPEGEOT: same here

Tests/Functional/Drivers/Queue/StompQueueDriverTest.php
moonraking: I think in this case we actaully do want to sleep for 1 sec, rather then mock sleep somewhere else

Tests/Functional/Drivers/Queue/StompQueueDriverTest.php
moonraking: Same here

Tests/Functional/Drivers/Queue/StompQueueDriverTest.php
moonraking: and here and also for the 3 sec sleep below

Tests/Functional/StepProviders/CsvConfigurableStepsProviderTest.php
moonraking: yes it would be nicer.
we would have to add it as a dependancy. Do we wana do that as part of this pull?

Tests/Functional/StepProviders/CsvConfigurableStepsProviderTest.php
moonraking: I have tidied these up now

Tests/Functional/Drivers/Queue/StompQueueDriverTest.php
GaryPEGEOT: In that case, ok.

Tests/Functional/StepProviders/CsvConfigurableStepsProviderTest.php
GaryPEGEOT: We require symfony/symfony (which is not especially a good idea) already, which include the FS component: https://packagist.org/packages/symfony/symfony#v2.8.49

PR #88, opened 2019-02-07
Update mapper to add functions to convert strings to SoapVars in an a…

melmccann:
…rray


GaryPEGEOT:
Can't see any test @melmccann, am I getting blind?

melmccann:
> Can't see any test @melmccann, am I getting blind?

@GaryPEGEOT Nope, not going blind... I will add one shortly

PR #89, opened 2019-02-07
Add the ability to override the base_uri with steps params

moonraking:


Tests/Unit/Components/WebService/Rest/RestConfigurableProducerTest.php
```diff
@@ -115,6 +134,99 @@ public function testItShouldCreateProperRestExceptions(
...
+            'exchange' => new Exchange(),
+        ];
+
+        //Decide to override the base_uri or not
+        if (false !== $overRideBaseUri) {
+            $actionParams['base_uri'] = $overRideBaseUri;
+        }
+
+        $response = $producer->executeStep(RestConfigurableProducer::STEP_REQUEST, $actionParams, $options, $context);
+        $this->assertTrue($response);
```
GaryPEGEOT: Please provide a meaningful message as second argument here, as 'Failing asserting that false is true' is not much help.

moonraking: added

PR #96, opened 2019-03-22
Feature: Drop connection on error

andreskrey:


Components/Queues/Drivers/StompQueueDriver.php
```diff
@@ -222,6 +223,14 @@ public function disconnect()
         }
     }
 
+    /**
+     * Kill the TCP connection directly
+     */
+    public function dropConnection()
```
moonraking: could this be protected instead?

PR #105, opened 2019-04-25
Add a way to handle consume exceptions

moonraking:


Components/Queues/Drivers/AmqpQueueDriver.php
```diff
@@ -241,9 +243,14 @@ public function receiveNoWait()
...
-        /** @var QueueMessageInterface $msg */
-        $msg = $this->getSerializer()->deserialize($this->currentEnvelope->getBody(), SerializableInterface::class, $this->format, $deserializationContext);
-
+        try{
+            /** @var QueueMessageInterface $msg */
+            $msg = $this->getSerializer()->deserialize($this->currentEnvelope->getBody(), SerializableInterface::class, $this->format, $deserializationContext);
+        } catch (\Exception $exception) {
+            $this->getExceptionHandler()($exception, $this->currentEnvelope->getBody());
+            $this->ack();
+            return null;
```
andreskrey: Shouldn't whatever is handling this exception be in charge of acking the message? Maybe we just want to log the error, tweak the message and continue processing it.

Components/Queues/Drivers/AmqpQueueDriver.php
```diff
@@ -241,9 +243,14 @@ public function receiveNoWait()
...
         $deserializationContext = new DeserializationContext();
 
-        /** @var QueueMessageInterface $msg */
-        $msg = $this->getSerializer()->deserialize($this->currentEnvelope->getBody(), SerializableInterface::class, $this->format, $deserializationContext);
-
+        try{
+            /** @var QueueMessageInterface $msg */
+            $msg = $this->getSerializer()->deserialize($this->currentEnvelope->getBody(), SerializableInterface::class, $this->format, $deserializationContext);
+        } catch (\Exception $exception) {
+            $this->getExceptionHandler()($exception, $this->currentEnvelope->getBody());
```
moonraking: Maybe we should pass the entire envelope here

PR #113, opened 2019-11-19
SF3.4 compatibility

andreskrey:


Resources/config/events_deferring.yml
```diff
@@ -3,11 +3,16 @@ parameters:
...
+  deferred_event_dispatcher:
+    class: "%event_dispatcher.class%"
+    arguments:
+      $container: '@service_container'
+
   smartesb.handlers.events:
       class: "%smartesb.handlers.events_deferring.class%"
       calls:
         - [setId, ['smartesb.handlers.events']]
-        - [setEventDispatcher, ['@event_dispatcher']]
```
campru: I have been investigating why in Symfony 2.8 this was working in our code and in Symfony 3.4 no and the reason is because in Symfony 2.8 `event_dispatcher.class` was defined as parameter and we were overriding the parameter here in framework and in Symfony 3.4 this parameter doesn't exist anymore and the event_dispatcher is declared directly as service.
This is the code in Symfony 2.8.47
symfony/symfony/src/Symfony/Bundle/FrameworkBundle/Resources/config/services.xml
```
<?xml version="1.0" ?>

<container xmlns="http://symfony.com/schema/dic/services"
    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
    xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">

    <parameters>
        <parameter key="event_dispatcher.class">Symfony\Component\EventDispatcher\ContainerAwareEventDispatcher</parameter>
        <parameter key="http_kernel.class">Symfony\Component\HttpKernel\DependencyInjection\ContainerAwareHttpKernel</parameter>
        <parameter key="filesystem.class">Symfony\Component\Filesystem\Filesystem</parameter>
        <parameter key="cache_warmer.class">Symfony\Component\HttpKernel\CacheWarmer\CacheWarmerAggregate</parameter>
        <parameter key="cache_clearer.class">Symfony\Component\HttpKernel\CacheClearer\ChainCacheClearer</parameter>
        <parameter key="file_locator.class">Symfony\Component\HttpKernel\Config\FileLocator</parameter>
        <parameter key="uri_signer.class">Symfony\Component\HttpKernel\UriSigner</parameter>
        <parameter key="request_stack.class">Symfony\Component\HttpFoundation\RequestStack</parameter>
    </parameters>

    <services>
        <service id="event_dispatcher" class="%event_dispatcher.class%">
            <argument type="service" id="service_container" />
        </service>
```
and this is the code in Symfony 3.4.35
```
<?xml version="1.0" ?>

<container xmlns="http://symfony.com/schema/dic/services"
    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
    xsi:schemaLocation="http://symfony.com/schema/dic/services https://symfony.com/schema/dic/services/services-1.0.xsd">

    <services>
        <defaults public="false" />

        <service id="event_dispatcher" class="Symfony\Component\EventDispatcher\ContainerAwareEventDispatcher" public="true">
            <argument type="service" id="service_container" />
            <tag name="container.hot_path" />
        </service>
        <service id="Symfony\Component\EventDispatcher\EventDispatcherInterface" alias="event_dispatcher" />
       ...
```
To be able to do the change in framework and don't affect to any of the other bundles, we should override the event_dispatcher in the compiler pass to define our class. https://symfony.com/doc/3.4/bundles/override.html#services-configuration

Declare the class names as parameters to allow other bundles override these classes was considered bad practice by symfony. See https://github.com/symfony/symfony/issues/11881

DependencyInjection/CompilerPasses/ExpressionLanguageCachePass.php
```diff
@@ -18,6 +18,10 @@ class ExpressionLanguageCachePass implements CompilerPassInterface
      */
     public function process(ContainerBuilder $container)
     {
+        // Last line (the reference to cache.app) triggers a "Constructing service "monolog.logger.cache" from a parent definition is not supported at build time."
+        // By bypassing this pass the container builds, but expression language throws a deprecation notice about not passing a cache interface in the constructor
+        return;
```
campru: Taking a look to this, ExpressionLanguage class has some special cases when defining the cache (https://github.com/symfony/symfony/blob/v3.4.35/src/Symfony/Component/ExpressionLanguage/ExpressionLanguage.php). This is implemented since symfony 3.2 (that is the deprecation you mentioned in the comment). What Gary did in that part was to add `cache.app` that is a service of `Symfony\Component\Cache\Adapter\ArrayAdapter` (see where these changes have been implemented https://github.com/smartboxgroup/integration-framework-bundle/pull/94/files) and this `cache.app` service has a call to `setLogger` to be able to log the exceptions (see file https://github.com/symfony/symfony/blob/v3.4.35/src/Symfony/Component/Cache/Traits/ArrayTrait.php). The default cache in `ExpressionLanguage` is `Symfony\Component\Cache\Adapter\ArrayAdapter` but without logger. Myabe we can try without logger to see if it works.

Tools/EventsDeferring/EventMessage.php
```diff
@@ -16,7 +16,7 @@ class EventMessage extends Message
     const HEADER_EVENT_NAME = 'event_name';
 
     /**
-     * @Assert\Valid(traverse=true, deep=true)
+     * @Assert\Valid(traverse=true)
```
campru: We should validate that this doesn't break in Symfony 2.8 but based on what I read I don't think it should affect because this property doesn't have arrays of objects. Here is some documentation I found comparing both https://stackoverflow.com/questions/29775773/symfony2-assert-valid-traverse-vs-deep

Resources/config/events_deferring.yml
```diff
@@ -3,6 +3,11 @@ parameters:
   event_dispatcher.class: Smartbox\Integration\FrameworkBundle\Tools\EventsDeferring\EventDispatcher
 
 services:
+  event_dispatcher:
```
campru: We need to validate that this is going to work in Symfony 2.8, that based on the documentation it should https://symfony.com/doc/2.8/service_container/service_decoration.html

Resources/config/events_deferring.yml
```diff
@@ -3,6 +3,11 @@ parameters:
...
+    class: "%event_dispatcher.class%"
```
campru: What we can do here is is replace ```class: "%event_dispatcher.class%"``` for ```class: Smartbox\Integration\FrameworkBundle\Tools\EventsDeferring\EventDispatcher``` because the mininum version of symfony we are supporting is 2.8 and defining the class as ```parameter``` was the way to override the service until symfony 2.6 (see https://symfony.com/doc/2.6/cookbook/bundles/override.html#services-configuration)

Resources/config/events_deferring.yml
campru: To override the service directly in the yaml file, we should copy the original definition of the service and just replace what we want (in this case the class) based on symfony 3.4 documentation (https://symfony.com/doc/3.4/service_container/service_decoration.html). Checking the **event_dispatcher** definition in Symfony 3.4, the service has a tag that is not included here (see below the event_dispatcher definition)
```
        <service id="event_dispatcher" class="Symfony\Component\EventDispatcher\ContainerAwareEventDispatcher" public="true">
            <argument type="service" id="service_container" />
            <tag name="container.hot_path" /> # Tag name
        </service>
```
This tag ```container.hot_path``` exists in Symfony 3.4 (https://symfony.com/doc/3.4/reference/dic_tags.html) but doesn't exist in Symfony 2.8
(https://symfony.com/doc/2.8/reference/dic_tags.html)
To be able have this service setup correctly in the symfony versions we support (2.8 and 3.4) I suggest to define a new compiler pass that overrides the class name of the service and leave the rest managed by the symfony version itself. This means that this service could be deleted from here, and also the parameter ```event_dispatcher.class: Smartbox\Integration\FrameworkBundle\Tools\EventsDeferring\EventDispatcher``` because this was the way to override services until Symfony 2.6 (see https://symfony.com/doc/2.6/cookbook/bundles/override.html#services-configuration)

Resources/config/events_deferring.yml
andreskrey: *.class parameters removed.

Tools/EventsDeferring/EventMessage.php
andreskrey: It is safe. It was already `true` in 2.8: https://github.com/symfony/validator/blob/2.8/Constraints/Valid.php#L30

Good point on removing `traverse`.

Resources/config/events_deferring.yml
andreskrey: The current way effectively overrides the event dispatcher in a backwards compatible way without a compiler pass.

DependencyInjection/CompilerPasses/ExpressionLanguageCachePass.php
campru: Did you take a look to this comment?

Resources/config/events_deferring.yml
campru: My point here was that in Symfony 3.4 this service has the tag name "**container.hot_path**" and with the override we are missing this information. This tag doesn't exist in symfony 2.8. For this reason the point of create a compiler pass

andreskrey: Seems that the tag is only used for debugging purposes: https://github.com/symfony/dependency-injection/blob/master/Dumper/PhpDumper.php#L153

I've added the tag in our definition, since it's backwards compatible with 2.8 (framework will ignore it)

DependencyInjection/CompilerPasses/ExpressionLanguageCachePass.php
campru: A separate ticket will be created to investigate this part. At the moment is having the same behaviour as with Symfony 2.8

PR #114, opened 2019-12-02
bugfix/manage-broken-utf8

jams246:
Fix handling of invalid UTF8 strings sent to the exception handler class.

### DIFF
--- a/Components/WebService/Rest/Exceptions/RestException.php
+++ b/Components/WebService/Rest/Exceptions/RestException.php
@@ -87,7 +87,7 @@ class RestException extends \Exception implements SerializableInterface, Externa
         $this->requestHttpHeaders = $requestHeaders;
         $this->requestHttpBody = $requestBody;
         $this->responseHttpHeaders = $responseHeaders;
-        $this->responseHttpBody = $responseBody;
+        $this->setResponseHttpBody($responseBody);
         $this->responseStatusCode = $responseStatusCode;
 
         $this->setOriginalCode($code);
@@ -172,6 +172,12 @@ class RestException extends \Exception implements SerializableInterface, Externa
      */
     public function setResponseHttpBody($responseHttpBody)
     {
+        $validUTF8 = !(false === mb_detect_encoding($responseHttpBody, 'UTF-8', true));
+        if (!$validUTF8) {
+            $responseHttpBody = utf8_decode($responseHttpBody);
+            $responseHttpBody = mb_convert_encoding($responseHttpBody, 'UTF-8', 'UTF-8');
+        }
+
         $this->responseHttpBody = $responseHttpBody;
     }
 
--- a/Tests/Functional/Exceptions/RestExceptionTest.php
+++ b/Tests/Functional/Exceptions/RestExceptionTest.php
@@ -1,0 +1,33 @@
+<?php
+
+
+namespace Smartbox\Integration\FrameworkBundle\Tests\Functional\Exceptions;
+
+use PHPUnit\Framework\TestCase;
+use Smartbox\Integration\FrameworkBundle\Components\WebService\Rest\Exceptions\RestException;
+
+class RestExceptionTest extends TestCase
+{
+    /**
+     * @param $string
+     * @param $expected
+     *
+     * @dataProvider provideExceptionData
+     */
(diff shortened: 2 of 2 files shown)


Components/WebService/Rest/Exceptions/RestException.php
```diff
@@ -172,6 +172,12 @@ public function getResponseHttpBody()
      */
     public function setResponseHttpBody($responseHttpBody)
     {
+        $validUTF8 = !(false === mb_detect_encoding($responseHttpBody, 'UTF-8', true));
```
campru: This place is one of the latest where we are dealing with the corrupt response. Checking the code of this class itself on the line https://github.com/smartboxgroup/integration-framework-bundle/pull/114/files#diff-f3a45c0d505d2b45322d0bca7e8844e6R94 we are going to get as response null because the json is broken.

Instead of catching the error almost at the end of the execution, we can catch this one in an early stage and throw and exception to treat with this scenario (when we are not able deserialise a payload). This is considered as best practise because we have the control of that particular issue and we can treat in the way we want. Otherwise the corrupt message goes to other parts of the code where it fails again and the exception and error message thrown are not the real cause of the error. In this case, the starting point of the workers dying is the following line https://github.com/smartboxgroup/integration-framework-bundle/blob/master/Components/WebService/Rest/RestConfigurableProducer.php#L218. In this one we are considering that if the message can't be deserialised we use the original message and we move forward. This provokes that later on the code (on the validation of the response) fails again because the response is not in the correct format and throw a new exception. For the system that fails, on the deserialise part has an option to deserialise to string type (the response type). Maybe the cause of having the truncated message is the type passed there. If that is not the reason, we should catch the RuntimeException and throw a new exception representing that we can't unserialise the response and later on catch this new exception to transform it to a recoverable exception.

andreskrey: With your approach the function remains vulnerable. If we use it somewhere else, we can trigger the same situation and end up killing the workers.

This class has the responsibility of sanitizing the data before storing it. Never trust user input.

campru: The function to **setResponseHttpBody** is to set the response, nothing else. If we need to sanitise the message it should be a new function for that and used in all the places where we are receiving input because all these places are potentially to kill the workers. The RestException class responsibility is to handle the exception, otherwise is losing the principle of having class with only 1 responsibility. We also have the **ExternalSystemHTTPEvent** that could have the same issue and we wouldn't start duplicating the sanitise method across all the possible places. 

My point is not if we need or not to sanitise the message, it is that we are allowing the corrupt message to go to far in the code and start throwing other exceptions with different error messages instead of catching when the real issue happens. We don't have control on that message and the reason why is failing because in the admin panel we will see a different error message.

We could catch the exception when not possible to deserialise, sanitise the message and throw the recoverable exception.

For the RestException itself, that function is not used anywhere in our code and in fact the arguments are passed in the constructor but the method is setup as public, so potentially someone could use.

andreskrey: I said class, not function.

jams246: > This place is one of the latest where we are dealing with the corrupt response. Checking the code of this class itself on the line https://github.com/smartboxgroup/integration-framework-bundle/pull/114/files#diff-f3a45c0d505d2b45322d0bca7e8844e6R94 we are going to get as response null because the json is broken.

null messages wont be captured in the $originalMessage property. The function will try to deserialize (as an associative array) the response received from the target system, If the deserialized object does not have the _message_ property, it will fallback to the value of $message; which in these cases will be the message from the thrown exception.

> The function to setResponseHttpBody is to set the response, nothing else.

A setter that does validation is acceptable since it's purpose is to maintain a valid state of the object. I'll move the sanitisation elsewhere.

campru: @andreskrey sorry for the part of class instead of function, but I don't think RestException class is the responsible of sanitising messages because if this sanitation needs to be used in other places (like ExternalSystemHTTPEvent) we would need to duplicate coode.

@jams246 my point was on the response of the json decode not on the ternary that exists later that is going to fallback to the value of $message

PR #116, opened 2019-12-13
WIP: Php amqp lib

campru:


Components/Queues/Drivers/AmqpQueueDriver.php
```diff
@@ -165,7 +165,7 @@ public function nack()
         $this->currentEnvelope = null;
     }
 
-    public function createQueueMessage()
+    public function createQueueMessage($queueName = null, $options = [])
```
campru: Which is the reason to add these arguments in a method that is not using them?

Components/Queues/QueueManager.php
```diff
@@ -133,7 +133,7 @@ public function getQueue(string $name, int $flag = AMQP_DURABLE, array $argument
 
     public function send(string $queueName, string $message, array $headers = [])
     {
-        $this->connect();
+//        $this->connect();
```
campru: Some particular reason of having this commented?

Components/Queues/QueueManager.php
```diff
@@ -147,10 +147,15 @@ public function send(string $queueName, string $message, array $headers = [])
...
 
-    public function addConnection(\AMQPConnection $connection)
+    public function addConnection($connection)
     {
         $this->connections[] = $connection;
 
         return $this;
     }
+    
+    public function getConnections()
```
campru: Is this new method used somewhere? I am not able to find any reference on the merge request

Components/Queues/QueueProducer.php
```diff
@@ -2,6 +2,7 @@
 
 namespace Smartbox\Integration\FrameworkBundle\Components\Queues;
 
+use Smartbox\Integration\FrameworkBundle\Components\Queues\Drivers\PhpAmqpLibDriver;
```
campru: This use is not needed here

Core/Endpoints/Endpoint.php
```diff
@@ -120,11 +120,14 @@ public function getProducer()
      */
     public function consume($maxAmount = 0)
     {
-        if ($maxAmount > 0) {
-            $this->getConsumer()->setExpirationCount($maxAmount);
+        try {
+            if ($maxAmount > 0) {
+                $this->getConsumer()->setExpirationCount($maxAmount);
+            }
+            $this->getConsumer()->consume($this);
```
campru: Which is the reason of adding this inside a try/catch?

Resources/config/consumers.yml
```diff
@@ -1,6 +1,8 @@
 parameters:
   smartesb.consumers.queue.class: Smartbox\Integration\FrameworkBundle\Components\Queues\QueueConsumer
   smartesb.consumers.async_queue.class: Smartbox\Integration\FrameworkBundle\Components\Queues\AsyncQueueConsumer
+  smartesb.consumers.php_amqp_lib_async_queue.class: Smartbox\Integration\FrameworkBundle\Components\Queues\PhpAmqpLibAsyncConsumer
+  smartesb.consumers.php_amqp_lib_signal_queue.class: Smartbox\Integration\FrameworkBundle\Components\Queues\PhpAmqpSignalConsumer
```
campru: Declare the class as parameter is considered a symfony bad practise, see https://github.com/symfony/symfony/issues/11881

Resources/config/consumers.yml
```diff
@@ -17,3 +19,19 @@ services:
           - [ setSmartesbHelper, ['@smartesb.helper']]
           - [ setEventDispatcher, ['@event_dispatcher']]
           - [ setSerializer, ['@jms_serializer']]
+  smartesb.consumers.php_amqp_lib_async_queue:
```
campru: Is still needed this class? I don't see is used in the code

Command/ConsumeCommand.php
```diff
@@ -98,11 +98,8 @@ protected function execute(InputInterface $input, OutputInterface $output)
         $message .= '.</info>';
         $output->writeln($message);
 
-        // AsyncQueueConsumer is immortal apart from Ctrl+C
-        if (!$this->endpoint->getConsumer() instanceof AsyncQueueConsumer) {
```
campru: Because you deleted this if, now shouldn't be needed the use reference to ```use Smartbox\Integration\FrameworkBundle\Components\Queues\AsyncQueueConsumer;```

Components/Queues/Drivers/PhpAmqpLibDriver.php
```diff
@@ -0,0 +1,385 @@
...
+
+/**
+ * Class PhpAmqpLibDriver
+ * @package Smartbox\Integration\FrameworkBundle\Components\Queues\Drivers
+ */
+class PhpAmqpLibDriver extends Service implements PurgeableQueueDriverInterface
+{
+    use UsesSerializer;
+    use UsesExceptionHandlerTrait;
+    use ContainerAwareTrait;
```
campru: Which is the reason this driver needs to be aware of the container?

Components/Queues/Drivers/PhpAmqpLibDriver.php
```diff
@@ -0,0 +1,385 @@
...
+
+    /**
+     * @var string Host form Message Broker
+     */
+    private $host;
+
+    /**
+     * @var string Username to connect to Message broker
+     */
+    private $username;
```
campru: I don't see any use of this property in this class. It could be deleted

Components/Queues/Drivers/PhpAmqpLibDriver.php
```diff
@@ -0,0 +1,385 @@
...
+
+    /**
+     * @var string $password
+     */
+    private $password;
```
campru: I don't see any use of this property in this class. It could be deleted

Components/Queues/Drivers/PhpAmqpLibDriver.php
```diff
@@ -0,0 +1,385 @@
...
+
+    /**
+     * @var string $port
+     */
+    private $port;
```
campru: I don't see any use of this property in this class. It could be deleted

Components/Queues/Drivers/PhpAmqpLibDriver.php
```diff
@@ -0,0 +1,385 @@
...
+
+    /**
+     * @var PhpAmqpLibQueueManager
+     */
+    private $manager;
+
+    /**
+     * @var string Host form Message Broker
+     */
+    private $host;
```
campru: I don't see any use of this property in this class. It could be deleted

Components/Queues/Drivers/PhpAmqpLibDriver.php
```diff
@@ -0,0 +1,385 @@
...
+
+    /**
+     * @var string $port
+     */
+    private $port;
+
+    /**
+     * @var string $vhost
+     */
+    private $vhost;
```
campru: I don't see any use of this property in this class. It could be deleted

Components/Queues/Drivers/PhpAmqpLibDriver.php
```diff
@@ -0,0 +1,385 @@
...
+
+    /**
+     * @var array
+     */
+    private $connectionId = [];
```
campru: I don't see any use of this property in this class. It could be deleted

Components/Queues/Drivers/PhpAmqpLibDriver.php
```diff
@@ -0,0 +1,385 @@
...
+
+    /**
+     * @var AMQPChannel
+     */
+    private $channel;
```
campru: I don't see any use of this property in this class. It could be deleted

Components/Queues/Drivers/PhpAmqpLibDriver.php
```diff
@@ -0,0 +1,385 @@
...
+
+    /**
+     * @param $host
+     * @param string $username
+     * @param string $password
+     * @param string $format
+     * @param null $port
+     * @param null $vhost
+     */
+    public function configure($host, $username, $password, $format = self::FORMAT_JSON, $port = null, $vhost = null)
```
campru: In the interface that defines this method ```Components/Queues/Drivers/QueueDriverInterface.php``` there is no the arguments ```port``` and ```vhost```

Components/Queues/Drivers/PhpAmqpLibDriver.php
```diff
@@ -0,0 +1,385 @@
...
+
+    /**
+     * PhpAmqpLibDriver constructor.
+     * @param QueueManager $manager
+     */
+    public function __construct(PhpAmqpLibQueueManager $manager)
+    {
+        parent::__construct();
+        $this->manager = $manager;
+        $this->format = self::FORMAT_JSON;
```
campru: Here it is assumed that the format this driver will manage the messages is json, but it could be xml or other format. This is defined in the ```configure``` method and then added on the ```DependencyInjection/SmartboxIntegrationFrameworkExtension.php``` based on what the user has defined in the driver configuration

Components/Queues/Drivers/PhpAmqpLibDriver.php
```diff
@@ -0,0 +1,385 @@
...
+
+    /**
+     * @param string $queue
+     * @throws \AMQPChannelException
+     * @throws \AMQPConnectionException
+     * @throws \AMQPQueueException
+     */
+    public function subscribe($queue)
+    {
+        if ($this->currentEnvelope) {
```
campru: You are using this property ```currentEnvelope``` but is not defined as property in the class

Components/Queues/Drivers/PhpAmqpLibDriver.php
```diff
@@ -0,0 +1,385 @@
...
+     *
+     * @throws \Exception
+     *
+     * @return \Smartbox\Integration\FrameworkBundle\Components\Queues\QueueMessageInterface|null
+     */
+    public function receiveNoWait()
+    {
+        if ($this->currentEnvelope) {
+            throw new \RuntimeException(
+                'AmqpQueueDriver: This driver has a message that was not acknowledged yet. A message must be processed and acknowledged before receiving new messages.'
```
campru: Here is mentioning the driver name of ```AmqpQueueDriver``` instead of ```PhpAmqpLibDriver```

Components/Queues/Drivers/PhpAmqpLibDriver.php
```diff
@@ -0,0 +1,385 @@
...
+        }
+    }
+
+    /**
+     *
+     */
+    private function fillByContainer()
+    {
+        $this->container->get();
+    }
```
campru: This method is not used in the class

Components/Queues/Drivers/PhpAmqpLibDriver.php
```diff
@@ -0,0 +1,385 @@
...
+     * @param string $queueName
+     * @param int $flag
+     * @param array $options
+     * @throws \AMQPChannelException
+     * @throws \AMQPConnectionException
+     * @throws \AMQPQueueException
+     */
+    public function setQueue($queueName)
+    {
+        $this->manager->declareQueue($queueName);
```
campru: I don't see any use of this method, where is it used?

Components/Queues/Drivers/PhpAmqpLibDriver.php
```diff
@@ -0,0 +1,385 @@
...
+    }
+
+    /**
+     * @param $exchange
+     */
+    public function setExchange($exchange)
+    {
+        $this->manager->declareExchange($exchange);
+    }
```
campru: I don't see any use of this method, where is it used?

Components/Queues/Handler/PhpAmqpHandler.php
```diff
@@ -0,0 +1,230 @@
...
+        } catch (\Exception $exception) {
+            $this->getExceptionHandler()($exception, ['headers' => $message->getHeaders(), 'body' => $message->getBody()]);
+        }
+    }
+
+    /**
+     * Set a AMQPChannel object to class variable
+     * @param AMQPChannel $channel
+     */
+    public function setChannel(AMQPChannel $channel): void
```
campru: The feature of marking the methods as return void was introduced in php 7.1 and framework bundle needs to support php 7.0

Components/Queues/Handler/PhpAmqpHandler.php
```diff
@@ -0,0 +1,230 @@
...
+/**
+ * Class PhpAmqpHandler
+ * @package Smartbox\Integration\FrameworkBundle\Components\Queues\Handler
+ */
+class PhpAmqpHandler implements LoggerAwareInterface
+{
+    use LoggerAwareTrait;
+    use UsesSmartesbHelper;
+    use UsesExceptionHandlerTrait;
+    use IsStopableConsumer;
```
campru: Which is the reason of this handler has the responsibility of stopping the consumer?

Components/Queues/Handler/PhpAmqpHandler.php
```diff
@@ -0,0 +1,230 @@
...
+        }
+    }
+
+    /**
+     * Verify if the channel is consuming a message.
+     * If there is message to consume it calls the consume callback function
+     * If there is no message to consume it will put the worker in a wait state
+     * @param AMQPChannel $channel
+     */
+    public function isConsuming(AMQPChannel $channel): void
```
campru: This method is only used inside the callback. It could be defined as private

Components/Queues/Handler/PhpAmqpHandler.php
```diff
@@ -0,0 +1,230 @@
...
+        $ctx['time'] = $now->format('Y-m-d H:i:s.u');
+        $this->logger->info($message, $ctx);
+    }
+
+    /**
+     * Dispatch the message to target system or events to database
+     * @param QueueMessage $message
+     * @throws \Exception
+     */
+    public function dispatchMessage(QueueMessage $message): void
```
campru: This method is only used inside the callback. It could be defined as private

Components/Queues/Handler/PhpAmqpHandler.php
```diff
@@ -0,0 +1,230 @@
...
+     * @param AMQPChannel $channel
+     * @param string $queueName
+     * @throws \Exception
+     */
+    public function consume(string $consumerTag, AMQPChannel $channel, string $queueName): void
+    {
+        $callback = function($message) use ($channel) {
+            $this->isConnected($channel);
+            // Send a message with the string "quit" to cancel the consumer.
+            if ($message->body === 'quit' || $this->shouldStop()) {
```
campru: The part of "Send a message with the string "quit" to cancel the consumer" is the implementation of the phpamqlib itself?

Components/Queues/PhpAmqpLibQueueManager.php
```diff
@@ -0,0 +1,309 @@
...
+            $this->connect();
+            $this->declareQueue($queueName);
+        }
+        return $this->queue ;
+    }
+
+    /**
+     * @return mixed
+     */
+    public function getId():? int
```
campru: This method doesn't seem to be used

Components/Queues/PhpAmqpLibQueueManager.php
```diff
@@ -0,0 +1,309 @@
...
+    {
+        return $this->id;
+    }
+
+    /**
+     * @param mixed $id
+     * @return PhpAmqpLibQueueManager
+     */
+    public function setId($id): self
```
campru: This method doesn't seem to be used

Components/Queues/PhpAmqpLibQueueManager.php
```diff
@@ -0,0 +1,309 @@
...
+     * PhpAmqpLibQueueManager constructor.
+     * @param \AMQPConnection[] $connections
+     */
+    public function __construct($connections = null, int $max = -1, string $format = 'json', SerializableInterface $serializer = null)
+    {
+        $this->connections = $connections;
+        $this->shoudlStop = false;
+        $this->max = $max;
+        $this->format = $format;
+        $this->serializer = $serializer ?? SerializerBuilder::create()->build();
```
campru: Part of what you have defined here it is also defined in the ```Components/Queues/Handler/PhpAmqpHandler.php``` like shouldStop, max, format, etc. Which class is the responsible of that?

Components/Queues/PhpAmqpLibQueueManager.php
```diff
@@ -0,0 +1,309 @@
...
+    }
+
+    /**
+     * Destroys the connection with the queuing system.
+     */
+    public function disconnect()
+    {
+        try {
+            if ($this->isConnected()) {
+                $this->connections[0]->close();
```
campru: If we are managing more than 1 connection, that is the case of EAI in production, we connect to 2 different HA proxies, not aways ```$this->connections[0]``` will be the one that will have the connection

Components/Queues/PhpAmqpLibQueueManager.php
```diff
@@ -0,0 +1,309 @@
...
+        }
+    }
+
+    /**
+     * Returns true if a connection already exists with the queueing system, false otherwise.
+     * @return bool
+     */
+    public function isConnected(): bool
+    {
+        return $this->connections[0]  ? $this->connections[0]->isConnected() : false;
```
campru: Here is the same, we have the chance to have different connections and not always [0] will be the connection used

Components/Queues/Drivers/PhpAmqpLibDriver.php
```diff
@@ -0,0 +1,385 @@
...
+
+    /**
+     * @param string $queueName
+     * @param int $flag
+     * @param array $options
+     * @throws \AMQPChannelException
+     * @throws \AMQPConnectionException
+     * @throws \AMQPQueueException
+     */
+    public function setQueue($queueName)
```
campru: I don't see any use of this method, where is it used?

Components/Queues/Drivers/PhpAmqpLibDriver.php
```diff
@@ -0,0 +1,385 @@
...
+    {
+        $this->manager->declareQueue($queueName);
+    }
+
+    /**
+     * @param $exchange
+     */
+    public function setExchange($exchange)
```
campru: I don't see any use of this method, where is it used?

Components/Queues/QueueManager.php
brunohsouza: Ok. I removed the function

Components/Queues/PhpAmqpLibQueueManager.php
```diff
@@ -0,0 +1,309 @@
...
+    {
+        $this->channel = reset($this->connections)->channel();
+        $this->channel->basic_qos(null, self::PREFETCH_COUNT, null);
+        return $this->channel;
+    }
+
+    /**
+     * AMQP Exchange is the publishing mechanism
+     */
+    public function declareExchange($exchange): void
```
campru: This method doesn't seem to be used in the code, there is only 1 reference in ```Components/Queues/Drivers/PhpAmqpLibDriver.php``` method ```setExchange``` but that method is not used anywhere

Components/Queues/PhpAmqpLibQueueManager.php
```diff
@@ -0,0 +1,309 @@
...
+        $this->connections[] = $connection;
+
+        return $this;
+    }
+
+    /**
+     * Returns all the connections available
+     * @return \AMQPConnection[]|array|null
+     */
+    public function getConnections():? array
```
campru: Where this method is used? I don't see any place where is used in the code

Components/Queues/PhpAmqpLibQueueManager.php
```diff
@@ -0,0 +1,309 @@
...
+    {
+        $this->id = $id;
+        return $this;
+    }
+
+    /**
+     * Shutdown the connection
+     * @throws \Exception
+     */
+    public function shutdown(): bool
```
campru: Which is the difference between shutdown and disconnect?

Components/Queues/Drivers/AmqpQueueDriver.php
brunohsouza: The reason is related to compatibility with some PHP versions like PHP 7.2. This versions requires the same method signature even if the parameters are not used. This was my other solution to remove that "if instance of PHPAMQPLib" from the QueueManager.php

Components/Queues/PhpAmqpSignalConsumer.php
```diff
@@ -0,0 +1,240 @@
...
+ * Class PhpAmqpSignalConsumer
+ * @package Smartbox\Integration\FrameworkBundle\Components\Queues
+ */
+class PhpAmqpSignalConsumer extends Service implements ConsumerInterface, LoggerAwareInterface
+{
+    use IsStopableConsumer;
+    use UsesSmartesbHelper;
+    use UsesSerializer;
+    use LoggerAwareTrait;
+    use UsesExceptionHandlerTrait;
```
campru: Are all these traits needed?

Components/Queues/QueueManager.php
brunohsouza: The function connect is called many times among the publish request. I was doing a test to see if I remove the call of this function here might cause some impact and there was none.

Components/Queues/QueueProducer.php
```diff
@@ -46,7 +47,7 @@ public function send(Exchange $ex, EndpointInterface $endpoint)
             throw new \RuntimeException("Found queue driver with name '$queueDriverName' that doesn't implement QueueDriverInterface");
         }
 
-        $queueMessage = $queueDriver->createQueueMessage();
+        $queueMessage = $queueDriver->createQueueMessage($queueName, $options);
```
campru: The Amqp native driver was creating the queueMessage without needed of ```$queueName``` and ```$options```. Why for the new driver is needed to pass these arguments?

DependencyInjection/SmartboxIntegrationFrameworkExtension.php
```diff
@@ -271,6 +273,49 @@ protected function loadQueueDrivers(ContainerBuilder $container)
                     }
                     break;
 
+                case 'phpamqplib':
+                    if (!class_exists(AMQPStreamConnection::class)) {
+                        throw new \RuntimeException('You need the amqp extension to use AMQP driver.');
```
campru: In this case, for the error message, we should mentioned that we are referring to phpAmqpLib

Components/Queues/QueueProducer.php
brunohsouza: Ok. I removed the use.

Core/Endpoints/Endpoint.php
brunohsouza: I was trying to catch an exception that happens when I call the consume function. But, as long as I've been treated this inside the consume function, I think it's not necessary anymore. So I will remove the try catch statements.

Resources/config/consumers.yml
brunohsouza: The goal to left this class is just in the case that as an Open Source lib, maybe the user don't wants to use the signal functions. So, after all it's just one more option to the lib.

DependencyInjection/SmartboxIntegrationFrameworkExtension.php
```diff
@@ -46,6 +47,7 @@ class SmartboxIntegrationFrameworkExtension extends Extension
         'smartesb.amqp.queue_manager',
         'smartesb.consumers.async_queue',
         'smartesb.drivers.queue.amqp',
+        'smartesb.consumers.php_amqp_lib_signal_queue.class'
```
campru: The purpose of this array was to remove the definitions of these services from the container. Based on this, all the new services you have implemented (manager, driver and consumers should be here)

campru:
On the composer json we should add as require-dev the phpamqp library we use

Resources/config/consumers.yml
brunohsouza: Ok. done.

Resources/config/consumers.yml
brunohsouza: Anyway, I will remove this class.

Command/ConsumeCommand.php
brunohsouza: Ok. I removed the useless use.

Components/Queues/Drivers/PhpAmqpLibDriver.php
brunohsouza: No needs. I removed the use of the Trait.

Components/Queues/Drivers/PhpAmqpLibDriver.php
brunohsouza: Ok. Removed with other useless variables

Components/Queues/Drivers/PhpAmqpLibDriver.php
brunohsouza: Ok. Removed with other useless variables

Components/Queues/Drivers/PhpAmqpLibDriver.php
brunohsouza: Ok. Removed with other useless variables

Components/Queues/Drivers/PhpAmqpLibDriver.php
brunohsouza: Ok. Removed with other useless variables

Components/Queues/Drivers/PhpAmqpLibDriver.php
brunohsouza: Ok. Removed with other useless variables

Components/Queues/Drivers/PhpAmqpLibDriver.php
brunohsouza: Ok. Removed with other useless variables

Components/Queues/Drivers/PhpAmqpLibDriver.php
brunohsouza: Ok. Removed with other useless variables

Components/Queues/Drivers/PhpAmqpLibDriver.php
brunohsouza: Ok, I removed the parameters once tested and verified that there is no impact int the functionality.

Components/Queues/Drivers/PhpAmqpLibDriver.php
brunohsouza: Yes. I altered this part to take the parameter based on configuration.

Components/Queues/Drivers/PhpAmqpLibDriver.php
brunohsouza: Yes, I have to implement the subscribe method. This class is here just because of the abstract parent class.

Components/Queues/Drivers/PhpAmqpLibDriver.php
brunohsouza: This function must to be customized for the new driver too. The reason that it's here is because of the parent abstract class too.

Components/Queues/Drivers/PhpAmqpLibDriver.php
brunohsouza: Ok, I removed.

Components/Queues/Drivers/PhpAmqpLibDriver.php
```diff
@@ -98,30 +75,32 @@ class PhpAmqpLibDriver extends Service implements PurgeableQueueDriverInterface
      * PhpAmqpLibDriver constructor.
      * @param QueueManager $manager
      */
-    public function __construct(PhpAmqpLibQueueManager $manager)
+    public function __construct(PhpAmqpLibQueueManager $manager, string $format = null)
```
campru: I see you add the format in the constructor, but I don't see the place where from the configuration this information is passed to the constructor. Based on that, it is always using json, but the format specified by the user in the driver config could be different than json

Components/Queues/Drivers/PhpAmqpLibDriver.php
```diff
@@ -98,30 +75,32 @@ class PhpAmqpLibDriver extends Service implements PurgeableQueueDriverInterface
...
         $this->manager = $manager;
-        $this->format = self::FORMAT_JSON;
+        $this->format = $format ?? self::FORMAT_JSON;
     }
 
     /**
      * PhpAmqpLibDriver destruct
      */
-    public function __destruct()
+    final function __destruct()
```
campru: Which is the reason behind to declare `__destruct()` as final? Also it is missing the visibility of the method (public, protected or private)

Components/Queues/Drivers/PhpAmqpLibDriver.php
```diff
@@ -149,8 +128,6 @@ public function createQueueMessage($queueName = null, $options = [])
     {
         $msg = new QueueMessage();
         $msg->setContext(new Context());
-        $this->setChannel();
-        $this->manager->declareQueue($queueName, AMQP_DURABLE, $options);
```
campru: Deleting this, you don't need anymore the `$queueName` and `$options` in the method definition

Components/Queues/Drivers/StompQueueDriver.php
```diff
@@ -386,7 +386,7 @@ public function nack()
         $this->currentFrame = null;
     }
 
-    public function createQueueMessage()
+    public function createQueueMessage($queueName = null, $options = [])
```
campru: The parameters `$queueName` and `$options` can be deleted from the method definition because they are not used

Components/Queues/Drivers/PhpAmqpLibDriver.php
```diff
@@ -0,0 +1,429 @@
...
+
+    /**
+     * PhpAmqpLibDriver destruct
+     */
+    final function __destruct()
+    {
+        try {
+            $this->manager->disconnect();
+        } catch (\Exception $connectionException) {
+            $this->getExceptionHandler()($connectionException);
```
campru: This piece of code is going to fail because the re is an extra parenthesis `getExceptionHandler()($connectionException)`
Another question is, what did it change in the manager disconnect method that we need to use the `getExceptionHandler`?

Components/Queues/Drivers/PhpAmqpLibDriver.php
```diff
@@ -0,0 +1,429 @@
...
+    /**
+     * @todo verify the need of this function
+     * {@inheritdoc}
+     */
+    public function doDestroy()
+    {
+        try {
+            $this->manager->disconnect();
+        } catch (\Exception $e) {
+            $this->getExceptionHandler()($e);
```
campru: The same comment I added before in the `__destruct()` method applies here

Components/Queues/Drivers/PhpAmqpLibDriver.php
```diff
@@ -0,0 +1,429 @@
...
+        }
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function send(QueueMessageInterface $queueMessage, $destination = null, $options = [])
```
campru: The `$options` parameter is not used inside the method

Components/Queues/Drivers/PhpAmqpLibDriver.php
```diff
@@ -0,0 +1,429 @@
...
+    {
+        $this->declareExchange($exchange);
+    }
+
+    /**
+     * Declares a channel to connect with the broker
+     */
+    public function setChannel()
+    {
+        $this->declareChannel();
```
campru: This call is not going to work because `declareChannel` requires some parameters. I also think that this method `setChannel` can be deleted based on the latest changes you did

Components/Queues/Drivers/PhpAmqpLibDriver.php
```diff
@@ -0,0 +1,429 @@
...
+     */
+    public function setExchange($exchange)
+    {
+        $this->declareExchange($exchange);
+    }
+
+    /**
+     * Declares a channel to connect with the broker
+     */
+    public function setChannel()
```
campru: Is this method used? I don't see any reference of this method

DependencyInjection/SmartboxIntegrationFrameworkExtension.php
```diff
@@ -301,7 +301,6 @@ protected function loadQueueDrivers(ContainerBuilder $container)
 
                     $driverDef = $container->findDefinition('smartesb.drivers.queue.phpamqplib');
                     $driverDef->addMethodCall('setId', [$driverId]);
-//                    $driverDef->addMethodCall('configure', [$driverConfig['host'], null, null, $driverConfig['format']]);
```
campru: This line should maintained here because from the driver yaml config the user can setup the format the of the data the driver is going to read and write to the queue

Components/Queues/Drivers/QueueDriverInterface.php
```diff
@@ -108,5 +108,5 @@ public function doDestroy();
     /**
      * @return int The time it took in ms to de-queue and deserialize the message
      */
-    public function getDequeueingTimeMs();
+    public function getDequeueTimeMs();
```
campru: Which is the reason of changing the interface signature? This provokes a breaking change in the bundle because this method is public and can be used for anybody

Components/Queues/AsyncQueueConsumer.php
```diff
@@ -39,7 +39,7 @@ class AsyncQueueConsumer extends Service implements ConsumerInterface, LoggerAwa
      * @param QueueManager $manager
      * @param string       $format
      */
-    public function __construct(QueueManager $manager, string $format = 'json')
+    public function __construct($manager, string $format = 'json')
```
campru: Is this consumer used with the phpAmqpLib? If not the type hinting QueueManager can be maintained in the code

Components/Queues/QueueManager.php
```diff
@@ -147,7 +147,7 @@ public function send(string $queueName, string $message, array $headers = [])
         return $this->exchange->publish($message, $queueName, AMQP_NOPARAM, $headers);
     }
 
-    public function addConnection(\AMQPConnection $connection)
+    public function addConnection($connection)
```
campru: Is this QueueManager used on your implementation?

Core/Endpoints/Endpoint.php
campru: It seems the try catch has not been removed yet

DependencyInjection/SmartboxIntegrationFrameworkExtension.php
```diff
@@ -271,6 +274,55 @@ protected function loadQueueDrivers(ContainerBuilder $container)
...
+                        throw new \RuntimeException('You need the amqp extension to use PHP-AMQP-LIB driver.');
+                    }
+
+                    if (empty($driverConfig['connections'] ?? [])) {
+                        throw new \InvalidArgumentException('You need to define at least one connection to use the AMQP driver.');
+                    }
+
+                    $driverDef = new Definition(PhpAmqpLibDriver::class);
+                    $driverDef->addMethodCall('setSerializer', [new Reference('jms_serializer')]);
+                    $driverDef->addMethodCall('setMessageFactory', [new Reference('smartesb.message_factory')]);
```
campru: These 2 objects can be passed directly in the service definition. In fact you are already doing that (see Resources/config/services.yml). For this reason this code
```
$driverDef->addMethodCall('setSerializer', [new Reference('jms_serializer')]);
$driverDef->addMethodCall('setMessageFactory', [new Reference('smartesb.message_factory')]);
```
can be deleted

DependencyInjection/SmartboxIntegrationFrameworkExtension.php
```diff
@@ -271,6 +274,55 @@ protected function loadQueueDrivers(ContainerBuilder $container)
...
+                            'port' => $connection['port'] ?? 5672,
+                            'user' => $connection['user'],
+                            'password' => $connection['pass'],
+                            'vhost' => trim($connection['path'] ?? '', '/'),
+                        ];
+                    }
+
+                    $driverDef->addMethodCall('setId', [$driverId]);
+                    $driverDef->addMethodCall('configure', [null, null, null, $driverConfig['format']]);
+                    $driverDef->addMethodCall('connect', [$connections]);
```
campru: Is at this point the connection stablished?

Components/Queues/Drivers/PhpAmqpLibDriver.php
```diff
@@ -0,0 +1,377 @@
...
+     * @throws \Exception
+     */
+    public function connect($connections = [], $shuffle = true)
+    {
+        try {
+            if ($shuffle) {
+                shuffle($connections);
+            }
+            $connection = AMQPStreamConnection::create_connection($connections, []);
+            $this->addConnection($connection);
```
campru: Can a specific process has more than 1 connection opened to rabbitMQ at the same time?

Components/Queues/Drivers/QueueDriverInterface.php
brunohsouza: This change was rollbackED

Components/Queues/AsyncQueueConsumer.php
brunohsouza: $manager param is typed again

Components/Queues/QueueManager.php
brunohsouza: Ok, $connection param is typed again.

DependencyInjection/SmartboxIntegrationFrameworkExtension.php
brunohsouza: These lines were removed.

DependencyInjection/SmartboxIntegrationFrameworkExtension.php
brunohsouza: No. The last call (to connect) is the start of the call to create the connections.

Components/Queues/Drivers/PhpAmqpLibDriver.php
brunohsouza: By the driver it must to return the first successful connection. By this way, there is just one connection per process. However, this array connection can contains more than one connection, but working always with the first one.

Components/Queues/Handler/PhpAmqpHandler.php
```diff
@@ -0,0 +1,232 @@
...
+                $channel->basic_cancel($message->delivery_info['consumer_tag']);
+                return false;
+            }
+
+            $this->log('A message was received on {time}');
+            $this->log('Message Body:' . $message->body);
+
+            $queueMessage = $this->deserializeMessage($message);
+            $this->dispatchMessage($queueMessage);
+            $channel->basic_ack($message->delivery_info['delivery_tag']);
```
andreskrey: I get that you're using the callback approach and calling the internal functions of the library directly, but basically you're bypassing the definition of [our own interface](https://github.com/smartboxgroup/integration-framework-bundle/blob/master/Components/Queues/Drivers/QueueDriverInterface.php).

The handler should implement all the functions that the interface defines and this callback should not bypass them.

andreskrey:
I've noticed that the connection timesout rather quickly. It happened to me that after a minute or two without sending messages (because I had a breakpoint somewhere else) the connection gets closed or an exception about "missed heartbeat" is thrown.

Wonder if there's a way to extend the timeout when the environment is `dev`. Otherwise you have to run all the time against this timeout (once the exception is thrown, the consumer stops)

Resources/config/services.yml
```diff
@@ -121,3 +121,11 @@ services:
...
       - [setMessageFactory, ['@smartesb.message_factory']]
     public: '%kernel.debug%'
+
+  # PHPAMQPLIB
+  smartesb.drivers.queue.phpamqplib:
+    class: Smartbox\Integration\FrameworkBundle\Components\Queues\Drivers\PhpAmqpLibDriver
+    calls:
+      - [setSerializer, ['@jms_serializer']]
+      - [setMessageFactory, ['@smartesb.message_factory']]
+    public: '%kernel.debug%'
```
andreskrey: Any reason to set it public based on the kernel env? In the AMQP one this is done to get it from the container directly when testing it.

Components/Queues/Handler/PhpAmqpHandler.php
```diff
@@ -0,0 +1,232 @@
...
+
+    /**
+     * Dispatch the message to target system or events to database
+     * @param QueueMessage $message
+     * @throws \Exception
+     */
+    private function dispatchMessage(QueueMessage $message)
+    {
+        if (!$this->endpoint) {
+            $this->getExceptionHandler()('Endpoint is undefined');
```
andreskrey: The exceptionHandler should only be used when you're dealing with incoming messages. Granted, the name is misleading, but it's main purpose is to requeue or rethrow the exception if the message fails to be deserialzed. 

If you don't declare it in your application, it rethrows the exception by default. In EAI, we override it with the `ReQueueSerializationExceptionHandler` which checks the type of exception and puts it in a different queue if the message is faulty.

In this case, and in most cases here, you just need to throw the exception. Only use this exception handler when you're deserializing the message.

Components/Queues/Handler/PhpAmqpHandler.php
```diff
@@ -0,0 +1,232 @@
...
+     * Deserialize the messages and dispatch them to the other targets
+     * @param AMQPMessage
+     * @return  MessageInterface|QueueMessageInterface $message
+     */
+    public function deserializeMessage(AMQPMessage $message): QueueMessageInterface
+    {
+        try {
+            return $this->serializer->deserialize($message->getBody(), SerializableInterface::class, $this->format);
+        } catch (\Exception $exception) {
+            $this->getExceptionHandler()($exception, ['headers' => $message->getHeaders(), 'body' => $message->getBody()]);
```
andreskrey: FYI, this is the only correct usage of the ExceptionHandler, every other call to it should be removed.

Components/Queues/Handler/PhpAmqpHandler.php
```diff
@@ -0,0 +1,232 @@
...
+    /**
+     * Consumes the message and dispatch it to others features
+     * @param string $consumerTag
+     * @param AMQPChannel $channel
+     * @param string $queueName
+     * @throws \Exception
+     */
+    public function consume(string $consumerTag, AMQPChannel $channel, string $queueName)
+    {
+        $callback = function($message) use ($channel) {
```
andreskrey: Would it be possible to move this away from this anon function and convert it to a real function? You could explicitly declare as a callback function if you do `return function(){ bla bla bla };` when saving it into a variable here.
I think it'd look cleaner that way.

Components/Queues/Handler/PhpAmqpHandler.php
```diff
@@ -0,0 +1,232 @@
...
+            $this->isConnected($channel);
+            // Send a message with the string "quit" to cancel the consumer.
+            if ($message->body === 'quit' || $this->shouldStop()) {
+                $channel->basic_cancel($message->delivery_info['consumer_tag']);
+                return false;
```
andreskrey: Any reason to return `false` here? Nothing seems to be using this boolean.

Core/Endpoints/Endpoint.php
brunohsouza: Sorry, it was removed now.

Components/Queues/Handler/PhpAmqpHandler.php
brunohsouza: All the return with type void was removed from the code.

Components/Queues/Handler/PhpAmqpHandler.php
brunohsouza: Done.

Components/Queues/Handler/PhpAmqpHandler.php
brunohsouza: Done.

Components/Queues/Handler/PhpAmqpHandler.php
brunohsouza: Yes, but it was removed.

Components/Queues/PhpAmqpLibQueueManager.php
brunohsouza: This class was removed.

Components/Queues/PhpAmqpLibQueueManager.php
brunohsouza: This class was removed.

Components/Queues/PhpAmqpLibQueueManager.php
brunohsouza: This class was removed.

Components/Queues/PhpAmqpLibQueueManager.php
brunohsouza: This class was removed.

Components/Queues/PhpAmqpLibQueueManager.php
brunohsouza: This class was removed.

Components/Queues/Drivers/PhpAmqpLibDriver.php
brunohsouza: This method was removed.

Components/Queues/Drivers/PhpAmqpLibDriver.php
brunohsouza: This class was removed.

Components/Queues/PhpAmqpLibQueueManager.php
brunohsouza: This method is not used inside the EAI scope because we don't use a specified exchange. But, I leave this method in case of need to declare some specific one.

Components/Queues/PhpAmqpLibQueueManager.php
brunohsouza: This class was removed.

Components/Queues/PhpAmqpLibQueueManager.php
brunohsouza: shutdown was removed

Components/Queues/PhpAmqpSignalConsumer.php
brunohsouza: The traits was validated. The unused ones was removed.

Components/Queues/QueueProducer.php
brunohsouza: It was changed to the former approach.

DependencyInjection/SmartboxIntegrationFrameworkExtension.php
brunohsouza: Ok, done.

DependencyInjection/SmartboxIntegrationFrameworkExtension.php
brunohsouza: Ok

brunohsouza:
> On the composer json we should add as require-dev the phpamqp library we use

It is on require-dev part.

Components/Queues/Drivers/PhpAmqpLibDriver.php
brunohsouza: Ok, it was changed.

Components/Queues/Drivers/PhpAmqpLibDriver.php
brunohsouza: IDE stuffs, I rolled back to the former.

Components/Queues/Drivers/PhpAmqpLibDriver.php
brunohsouza: Done.

Components/Queues/Drivers/StompQueueDriver.php
brunohsouza: Done.

Components/Queues/Drivers/PhpAmqpLibDriver.php
brunohsouza: It was modified.

Components/Queues/Drivers/PhpAmqpLibDriver.php
brunohsouza: It was modified.

Components/Queues/Drivers/PhpAmqpLibDriver.php
brunohsouza: Done

Components/Queues/Drivers/PhpAmqpLibDriver.php
brunohsouza: It was removed

Components/Queues/Drivers/PhpAmqpLibDriver.php
brunohsouza: it was removed

DependencyInjection/SmartboxIntegrationFrameworkExtension.php
brunohsouza: Ok, it was modified.

DependencyInjection/SmartboxIntegrationFrameworkExtension.php
brunohsouza: It was modified.

Components/Queues/Drivers/PhpAmqpLibDriver.php
```diff
@@ -0,0 +1,377 @@
...
+     * Method responsible to open the connection with the broker
+     *
+     * @param array $connections
+     * @param bool $shuffle - used to randomize the connections if exists more than one
+     * @throws \Exception
+     */
+    public function connect($connections = [], $shuffle = true)
+    {
+        try {
+            if ($shuffle) {
```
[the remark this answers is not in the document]
brunohsouza: As we discuss on the meeting room. The shuffle happens every time the driver does a new connection and there are 2 or more hosts defined. In other cases, the request should to fall on the function tovalidate if there is or not an existent connection.

Components/Queues/Drivers/PhpAmqpLibDriver.php
```diff
@@ -0,0 +1,377 @@
...
+                shuffle($connections);
+            }
+            $connection = AMQPStreamConnection::create_connection($connections, []);
+            $this->addConnection($connection);
+        } catch (\AMQPConnectionException $connectionException) {
+            echo $connectionException->getMessage();
```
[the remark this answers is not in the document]
brunohsouza: Every echoes was removed from the code and replaced for an exception handler call.

Components/Queues/Handler/PhpAmqpHandler.php
```diff
@@ -0,0 +1,232 @@
...
+
+    /**
+     * Log in console the message passed as param
+     * @param string $message
+     * @param array $ctx
+     * @throws \Exception
+     */
+    private function log(string $message, array $ctx = [])
+    {
+        if (null === $this->logger) {
```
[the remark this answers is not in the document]
brunohsouza: The log will display the message on the console in runtime. It's exactrly how the other drivers on this bundle works.

Components/Queues/Drivers/PhpAmqpLibDriver.php
```diff
@@ -0,0 +1,377 @@
...
+     * @param array $options
+     */
+    public function publish(AMQPMessage $message, string $queueName, array $options = [])
+    {
+        try {
+            $this->declareChannel();
+            $this->declareQueue($queueName, AMQP_DURABLE, $options);
+            $this->channel->basic_publish($message, self::EXCHANGE_NAME, $queueName);
+        } catch (\Exception $exception) {
+            $this->getExceptionHandler()($exception, ['headers' => $message->getHeaders(), 'body' => $message->getBody()]);
```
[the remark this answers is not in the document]
brunohsouza: It was modified to catch the exception message.

Components/Queues/Handler/PhpAmqpHandler.php
brunohsouza: This approach will be implemented on the future, maybe. The first idea was just to validate the driver and the features that we can do with it.

brunohsouza:
> I've noticed that the connection timesout rather quickly. It happened to me that after a minute or two without sending messages (because I had a breakpoint somewhere else) the connection gets closed or an exception about "missed heartbeat" is thrown.
> 
> Wonder if there's a way to extend the timeout when the environment is `dev`. Otherwise you have to run all the time against this timeout (once the exception is thrown, the consumer stops)

The timesout was increased to 60 seconds each call.

Resources/config/services.yml
brunohsouza: No reason. I was just following the implementations for the old driver

Components/Queues/Handler/PhpAmqpHandler.php
brunohsouza: Ok, It's working with the exceptionHandler, but I'll refactory this soon.

Components/Queues/Handler/PhpAmqpHandler.php
brunohsouza: Alright, see the comment above.

Components/Queues/Handler/PhpAmqpHandler.php
brunohsouza: No, I tried, but to use the callback inside the method it should be on the same scope. Otherwise, it will not be able to access.

Components/Queues/Handler/PhpAmqpHandler.php
brunohsouza: Yes, to inform that the consume fails and stop the execution.

andreskrey:
Moved to https://github.com/smartboxgroup/integration-framework-bundle/pull/117

PR #119, opened 2020-03-13
feature/add-guzzle-handler-for-http-errors: Adding exception, middlew…

andreymor:
…are and client


docker-compose.yml
```diff
@@ -15,4 +15,4 @@ services:
       - RABBITMQ_DEFAULT_USER=guest
       - RABBITMQ_DEFAULT_PASS=guest
     ports:
-      - "8081:15672"
\ No newline at end of file
+      - "8087:15672"
```
andreskrey: hwat?

Components/WebService/Rest/Middleware.php
```diff
@@ -0,0 +1,76 @@
...
+        $size = $body->getSize();
+
+        if (0 === $size) {
+            return null;
+        }
+
+        $summary = $body->read($size);
+
+        if (self::$truncateResponseSize > 0 && $size > self::$truncateResponseSize) {
+            if (false === mb_detect_encoding($summary, 'UTF-8', true)) {
```
andreskrey: Why do you want to validate if the content is in UTF-8? I don't like the idea of converting the response to something else... It might end up hiding the real error in the request. Plus `mb_substr` should be smart enough to truncate the string safely.

Components/WebService/Rest/Middleware.php
```diff
@@ -0,0 +1,76 @@
...
+            return null;
+        }
+
+        $size = $body->getSize();
+
+        if (0 === $size) {
+            return null;
+        }
+
+        $summary = $body->read($size);
```
andreskrey: Comparing with the original class, here you're forgetting to `rewind()` the `$body` like this

```
        $body->rewind();
```

Components/WebService/Rest/RestConfigurableProducer.php
```diff
@@ -202,6 +202,12 @@ protected function request(ClientInterface $client, array $stepActionParams, arr
         $request = new Request($httpMethod, $resolvedURI, $requestHeaders);
         $response = null;
         try {
+
+            //setting the new error_handler
+            if(null !== $handlerStack = $client->getConfig('handler')) {
+                $handlerStack->push(Middleware::httpErrors(), 'http_errors_handler');
```
andreskrey: We might need to move this step to the container or setup of the client, but not here. If you run the worker with a couple of messages going to the same producer, you end up stacking the same handler over and over and over. This is what I got after processing two messages that failed to be sent to the same target (`var_export()` of the handler stack):

```
array (
  0 => 
  array (
    0 => 
    Closure::__set_state(array(
    )),
    1 => 'http_errors',
  ),
  1 => 
  array (
    0 => 
    Closure::__set_state(array(
    )),
    1 => 'allow_redirects',
  ),
  2 => 
  array (
    0 => 
    Closure::__set_state(array(
    )),
    1 => 'cookies',
  ),
  3 => 
  array (
    0 => 
    Closure::__set_state(array(
    )),
    1 => 'prepare_body',
  ),
  4 => 
  array (
    0 => 
    Closure::__set_state(array(
    )),
    1 => 'http_errors_handler',
  ),
  5 => 
  array (
    0 => 
    Closure::__set_state(array(
    )),
    1 => 'http_errors_handler',
  ),
)
```

Notice the two `http_errors_handler` at the bottom of the array.

Tests/Unit/Components/WebService/Rest/MiddlewareTest.php
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    public function provideResponseData()
+    {
+        $response = "Relationship type 'Experience-Component' cannot be applied on product type 'Experience' of child product 'Chasse au trésor au Centre Pompidou'.";
+        $fullResponse = "Relationship type 'Experience-Component' cannot be applied on product type 'Experience' of child product 'Chasse au trésor au Centre Pompidou'.";
+        $defaultTruncate = "Relationship type 'Experience-Component' cannot be applied on product type 'Experience' of child product 'Chasse au trés (truncated...)";
+        $truncatedAt135 = "Relationship type 'Experience-Component' cannot be applied on product type 'Experience' of child product 'Chasse au trésor au Centre Po (truncated...)";
+
+        $encodedResponse = utf8_decode($response);
```
andreskrey: This is messing up the encoding of the original response:

`Relationship type 'Experience-Component' cannot be applied on product type 'Experience' of child product 'Chasse au tr�sor au Centre Pompidou'.`

andreymor: This was for the previous version that had UTF-8 check.
Its already removed.

Components/WebService/Rest/RestConfigurableProducer.php
andreymor: Moved to UseGuzzeTrait

Components/WebService/Rest/Middleware.php
andreymor: There is no need to rewind the body.

Components/WebService/Rest/Middleware.php
andreymor: Removed

DependencyInjection/SmartboxIntegrationFrameworkExtension.php
```diff
@@ -108,6 +109,11 @@ public function loadConfigurableProducers(ContainerBuilder $container)
             $definition->addMethodCall('setSerializer', [new Reference('jms_serializer')]);
             $definition->addMethodCall('setEventDispatcher', [new Reference('event_dispatcher')]);
             $definition->addMethodCall('setName', [$producerName]);
+
+            if(is_subclass_of($class, RestConfigurableClientInterface::class)) {
```
campru: Can't be use here is **instanceof**?

andreymor: We can if we use an instance (new $class) instanceOf to compare objects.
Comparing string with FCQN does not seems to work properly

`
$interfaceName = RestConfigurableClientInterface::class;

            if(is_subclass_of($class, $interfaceName)) //true
         
            if($class instanceof $interfaceName) //false`

campru: With the instance of you need to specify the class itself, not the class name, maybe for this reason doesn't work well

andreymor: Thats just what I said at the first statement.
if we use an instance (new $class) instanceOf to compare objects.
Or is an object comparison or the usage of subclass
$class (its a FCQN) instanceOf Interface (is an object) does not work

campru: Thanks for the clarification :)

As bonus track, can you run phpcs on the changes you did?

andreymor: Done!

andreskrey:
GTG

PR #120, opened 2020-04-02
Params/amqp

brunohsouza:
Adding new parameters to queue drivers.

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -397,7 +397,7 @@ class Configuration implements ConfigurationInterface
             ->defaultValue(null)
             ->end()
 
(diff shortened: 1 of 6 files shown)


DependencyInjection/Configuration.php
```diff
@@ -393,14 +393,22 @@ public function addQueueDriversNode()
             ->isRequired()
             ->end()
 
-            ->scalarNode('timeout')
+            ->scalarNode('read_timeout')
+            ->defaultValue(null)
+            ->end()
+
+            ->scalarNode('connection_timeout')
```
andreskrey, line 400: Isn't this a float? I would use `floatNode` here instead.

PR #121, opened 2020-04-06
Feature/exception handler transparent for the driver

campru:


Components/Queues/AsyncQueueConsumer.php
```diff
@@ -137,26 +128,39 @@ protected function process(EndpointInterface $queueEndpoint, MessageInterface $m
...
      *
      * {@inheritdoc}
-     *
-     * @return callable
      */
     protected function callback(EndpointInterface $endpoint): callable
     {
-        return function (AMQPMessage $message) use ($endpoint) {
+        return function (AMQPMessage $encodedMessage) use ($endpoint) {
+            $driver = $this->getQueueDriver($endpoint);
```
andreskrey, line 135: Can't we move this assignment inside the `catch`? It's a call to the container that might end up not being used.

campru: Yes, definitely, it was there because before was also used on the `try` part but now is not anymore and I missed to move. Good catch

Core/Consumers/AbstractConsumer.php
```diff
@@ -20,15 +22,18 @@ abstract class AbstractConsumer extends Service implements ConsumerInterface
     use UsesSmartesbHelper;
 
     /**
-     * Initializes the consumer for a given endpoint.
+     * Hold the amount of time it took to consume a message.
```
andreskrey, line 25:
```suggestion
     * Holds the amount of time it took to consume a message.
```

Tools/SmokeTests/QueueDriverConnectionSmokeTest.php
```diff
@@ -39,13 +38,9 @@ public function run()
                 throw new \RuntimeException('Function isConnected() returned false.');
             }
 
-            $message = new QueueMessage();
-            $message->setTTL(1);
-            $message->setQueue('isAlive');
-
-            if (!$this->queueDriver->send($message)) {
+            if (!$this->queueDriver->send('isalive')) {
```
andreskrey, line 41:
```suggestion
            if (!$this->queueDriver->send('isalive', '', ['expiration' => 1000])) {
```

andreskrey: This will work for AMQP, not sure for STOMP. If the header is different for STOMP, we can just add it as another key of the same array.

campru: I removed the "ttl" (internally is also the expieration) because what was happening was that the message after 1 second was going to the **dlq**. Don't remember with which driver this was happening. I will check in both drivers what happens

campru: After checking what was happening was the following. I had the **isalive** queue with the policy of the **DLX** and for this reason the messages were going to the **dlq** after 1 second. When I applied the same setup as production, we have a policy called **isalive** that is applied to the **isalive** queue and add expiration of the messages to 1 second, the messages where expiring and disappear from the queue as we expect. For this reason I removed the expiration header and it works for both drivers

I think in this case should be the user deciding what do to with this message through a policy in the broker rather than ourself defining a ttl in the code

Core/Consumers/AbstractConsumer.php
campru: Changed

Components/Queues/AsyncQueueConsumer.php
campru: Changed

Components/Queues/AsyncQueueConsumer.php
```diff
@@ -137,26 +128,36 @@ protected function process(EndpointInterface $queueEndpoint, MessageInterface $m
...
+                $this->getExceptionHandler()->handle($exception, [
+                    'endpoint' => $endpoint,
+                    'body' => $encodedMessage->getBody(),
+                    'headers' => $encodedMessage->get('application_headers')->getNativeData(),
+                ]);
+                $message = new QueueMessage();
+                $message->setMessageId($encodedMessage->getDeliveryTag());
+                $this->getQueueDriver($endpoint)->ack($message);
+
                 $this->consumptionDuration = (microtime(true) - $start) * 1000;
```
andreskrey, line 155: What about moving all this functionality to its own separate function, like `finalizeProcessing()` or something more catchy? In the end, we want this chunk to emulate what the parent does, so we are repeating ourselves.

You're also forgetting here to reset the `consumptionDuration` property. (parent does it after dispatching the vent)

Components/Queues/AsyncQueueConsumer.php
```diff
@@ -137,26 +128,36 @@ protected function process(EndpointInterface $queueEndpoint, MessageInterface $m
...
-                $this->getExceptionHandler()($exception, ['message' => $message]);
 
-                return false;
+                return null;
```
andreskrey, line 157: Would it make sense to expect something from the exception handler, and if returns `null`, consider the message poisoned and `ack` it? Otherwise it doesnt make sense to have this functionality when the built-in handler just rethorws the exception (and in essence, prevents this code from being executed).

Something like this (pseudocode)
```
if($message = handler->handle() === null) {
    ack()
    return null
}

parent::callback()
```

andreskrey: Or even better (maybe it's an overkill), but trigger the handler wrapped in a `try/catch` that, if it catches a `PosionedMessageException`, acks it, otherwise continues with the execution.

campru: Changed to the option one

Components/Queues/AsyncQueueConsumer.php
campru: The `consumptionDuration` is not needed to be reset because I am using an assignation `=`

About the other part could be cool to review all the duplicated coded between AbstractConsumer, AbstractAsyncConsumer, QueueConsumer and AsyncQueueConsumer. Not sure if we will have the time now

DependencyInjection/Traits/UsesQueueSerializer.php
```diff
@@ -0,0 +1,25 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Smartbox\Integration\FrameworkBundle\DependencyInjection\Traits;
+
+use Smartbox\Integration\FrameworkBundle\Core\Serializers\QueueSerializerInterface;
```
andreskrey, line 7: What's the purpose of this `use` statement here? It won't force you to implement what the interface defines.

Components/Queues/AsyncQueueConsumer.php
```diff
@@ -137,26 +126,42 @@ protected function process(EndpointInterface $queueEndpoint, MessageInterface $m
...
             } catch (\Exception $exception) {
-                $this->consumptionDuration = (microtime(true) - $start) * 1000;
-                $this->getExceptionHandler()($exception, ['message' => $message]);
+                $message = $this->getDecodingExceptionHandler()->handle($exception, [
+                    'endpoint' => $endpoint,
+                    'body' => $body,
+                    'headers' => $headers,
+                ]);
+
+                if (null === $message) {
```
andreskrey, line 149:
```suggestion
                // If the exception handler doesn't return a new message, consider it poisoned and discard it
                if (null === $message) {
```

DependencyInjection/Traits/UsesQueueSerializer.php
campru: In this case is forcing to set a serializer that implements that interface

andreskrey: Nvm i didnt see the setter type hinting

Components/Queues/Serialization/Serializer.php
```diff
@@ -0,0 +1,59 @@
...
+     */
+    public function decode(array $encodedMessage): MessageInterface
+    {
+        if (empty($encodedMessage['body']) || !is_array($encodedMessage['headers'])) {
+            throw new MessageDecodingFailedException('Encoded message should have at least a "body" and some "headers"');
+        }
+
+        try {
+            $message = $this->serializer->deserialize($encodedMessage['body'], SerializableInterface::class, $this->format);
+        } catch (RuntimeException $e) {
```
andreskrey, line 38: Would it make sense to catch the `InvalidArgumentException` from doctrine here as well? Currently this one just rethrows another one, and on EAI's side we check for doctrine's as well... Since this is thrown by JSM, it might be wise to catch it here as well.

andreskrey: This would simplify EAI's side as well, as it only needs to check for one exception only.

Tools/SmokeTests/QueueDriverConnectionSmokeTest.php
andreskrey: By definition a health check message should be short lived, either by queue policy or by message TTL. Given that framework doesn't provide any policies, nor forces you to use any (retry/throttle policies are "dumb" in framework, since they just nack the message when they are not ready to be processed), I'd say it's the smoke test function responsibility to make it short lived. If it goes to a DLX because of the users' policy, the user needs to accommodate that policy to what the framework expects. In my opinion!

Components/Queues/AsyncQueueConsumer.php
campru: Comment added in both consumers

Tools/SmokeTests/QueueDriverConnectionSmokeTest.php
campru: Changed

Components/Queues/Serialization/Serializer.php
campru: Added

Tools/SmokeTests/QueueDriverConnectionSmokeTest.php
andreskrey: Changed to x-message-ttl to make it compatible with our current queue policy.

campru: Cool!

andreskrey:
GTG

PR #125, opened 2020-04-20
Fix issues detected by scrutinizer in Amqp and Stomp queue drivers

campru:


Components/Queues/Drivers/PhpAmqpLibDriver.php
```diff
@@ -316,7 +316,8 @@ public function send(QueueMessageInterface $message, $destination = null): bool
      */
     public function declareQueue(string $queueName, int $durable = QueueMessage::DELIVERY_MODE_PERSISTENT, array $arguments = [])
```
campru, line 317: For the durable part, I would change the signature to boolean because persistent and durable are not directly connected. You can have in a durable queue a message not persistent. Based on that my suggestion is the following
```
public function declareQueue(string $queueName, boolean $durable = true, array $arguments = [])
{
    return $this->channel->queue_declare($queueName, false, $durable, false, false, false, new AMQPTable($arguments));
}
```

campru: And then with this we need to change inside the `send()` of `AmqpQueueDriver` to 
```
public function send(string $destination, string $body = '', array $headers = []): bool
    {
        $this->declareChannel();
        $this->declareQueue(
            $destination,
            true,
            array_filter($headers, function ($value) {
                return 0 === strpos($value, 'x-');
            }, ARRAY_FILTER_USE_KEY));

```

brunohsouza: Done.

PR #128, opened 2020-04-30
New response_format and request_fallback parameters

andreskrey:
Two new parameters:
* response_format parameter that allows to pass custom deserializers to jms.
* request_fallback that allows to fallback into the original response when framework fails to deserialize it (deafults to true)


Components/WebService/Rest/RestConfigurableProducer.php
```diff
@@ -112,6 +113,7 @@ protected function request(ClientInterface $client, array $stepActionParams, arr
             self::VALIDATION,
             self::REQUEST_QUERY_PARAMETERS,
             self::REQUEST_BASE_URI,
+            self::REQUEST_EXPECTED_RESPONSE_FORMAT
```
campru, line 116: Couldn't we define the `REQUEST_EXPECTED_RESPONSE_FORMAT` in the same way as `REQUEST_EXPECTED_RESPONSE_TYPE`? I mean
```
$stepParamsResolver->setDefault(self::REQUEST_EXPECTED_RESPONSE_FORMAT, $encoding);
$stepParamsResolver->setDefined([
            RestConfigurableProtocol::OPTION_HEADERS,
            self::VALIDATION,
            self::REQUEST_QUERY_PARAMETERS,
            self::REQUEST_BASE_URI,
])
...
$responseBody = $this->getSerializer()->deserialize(
    $responseContent,
    $params[self::REQUEST_EXPECTED_RESPONSE_TYPE],
    $params[self:: REQUEST_EXPECTED_RESPONSE_FORMAT],
);
```

Components/WebService/Rest/RestConfigurableProtocol.php
```diff
@@ -15,6 +15,8 @@ class RestConfigurableProtocol extends ConfigurableWebserviceProtocol implements
     const OPTION_HEADERS = 'headers';
     const OPTION_AUTH = 'authentication';
     const OPTION_ENCODING = 'encoding';
+    /** @deprecated: Will be removed on next major version. Fallback will not happen in v3. */
+    const OPTION_REQUEST_FALLBACK_ON_ERROR = 'request_fallback';
```
campru, line 19: In this case I think the value of the constant to be more accurate should be `response_fallback` because we are not doing any fallback with the request
Also based on that I would rename the constant to `OPTION_RESPONSE_FALLBACK_ON_ERROR'`

Components/WebService/Rest/RestConfigurableProtocol.php
```diff
@@ -15,6 +15,8 @@ class RestConfigurableProtocol extends ConfigurableWebserviceProtocol implements
     const OPTION_HEADERS = 'headers';
     const OPTION_AUTH = 'authentication';
     const OPTION_ENCODING = 'encoding';
+    /** @deprecated: Will be removed on next major version. Fallback will not happen in v3. */
```
campru, line 18:
```suggestion
    /** @deprecated: It will be removed on next major version. Fallback will not happen in v3. */
```

Components/WebService/Rest/RestConfigurableProtocol.php
```diff
@@ -58,6 +60,7 @@ public function configureOptionsResolver(OptionsResolver $resolver)
             self::OPTION_HEADERS => [],
             self::OPTION_AUTH => '',
             self::OPTION_ENCODING => self::ENCODING_JSON,
+            self::OPTION_REQUEST_FALLBACK_ON_ERROR => true,
```
campru, line 63: On the part below of this method we should define that the allow type for `OPTION_REQUEST_FALLBACK_ON_ERROR` should be `'boolean'`

Components/WebService/Rest/RestConfigurableProducer.php
```diff
@@ -199,14 +201,19 @@ protected function request(ClientInterface $client, array $stepActionParams, arr
...
+                        $params[self::REQUEST_EXPECTED_RESPONSE_FORMAT] ?? $encoding
                     );
                 } catch (RuntimeException $e) {
                     // Assume that the exception is one of the JSON exceptions that will kill the workers
                     if (RestConfigurableProtocol::ENCODING_JSON === $encoding && (JSON_ERROR_SYNTAX != json_last_error())) {
                         throw new UnexpectedValueException($e->getMessage());
                     }
-                    // if it cannot parse the response fallback to the textual content of the body
+
+                    if (!$endpointOptions[RestConfigurableProtocol::OPTION_REQUEST_FALLBACK_ON_ERROR]) {
```
campru, line 212: Because we know the only options for this can be `true` or `false`, we could define the condition like this
```
if (false === $endpointOptions[RestConfigurableProtocol::OPTION_RESPONSE_FALLBACK_ON_ERROR])
```

Components/WebService/Rest/RestConfigurableProducer.php
```diff
@@ -199,14 +201,19 @@ protected function request(ClientInterface $client, array $stepActionParams, arr
...
+                        throw $e;
+                    }
+
+                    trigger_error('Falling back to the original payload when the deserialization failed and a custom response format isn\'t defined will throw an exception in v3.', E_USER_DEPRECATED);
```
campru, line 216:
```suggestion
                    trigger_error('Falling back to the original response payload when the deserialization fails and a custom response format isn\'t defined will throw an exception in v3.', E_USER_DEPRECATED);
```

Components/WebService/Rest/RestConfigurableProducer.php
andreskrey: I agree that it would be more elegant but `$encoding` is calculated later on this monstrous method. Getting `$encoding` before this chunk would require rearranging the code a bit and it might have some side effects.

Components/WebService/Rest/RestConfigurableProtocol.php
andreskrey: Agreed. Updated on both ends.

Components/WebService/Rest/RestConfigurableProtocol.php
andreskrey: It's a null subject phrase in imperative form. There's no need for a subject: https://en.wikipedia.org/wiki/Null-subject_language#The_imperative_form

Components/WebService/Rest/RestConfigurableProducer.php
campru: Fair enough

Components/WebService/Rest/RestConfigurableProtocol.php
campru: Ok, when reading the first time I took as informative sentence rather than imperative

Components/WebService/Rest/RestConfigurableProtocol.php
andreskrey: Good idea, added.

Components/WebService/Rest/RestConfigurableProducer.php
andreskrey: Done

andreskrey:
All comments addressed

Components/WebService/Rest/RestConfigurableProtocol.php
```diff
@@ -72,6 +75,7 @@ public function configureOptionsResolver(OptionsResolver $resolver)
         $resolver->setAllowedTypes(self::OPTION_AUTH, ['string', 'null']);
         $resolver->setAllowedTypes(self::OPTION_ENCODING, ['string']);
         $resolver->setAllowedValues(self::OPTION_ENCODING, [self::ENCODING_JSON, self::ENCODING_XML]);
+        $resolver->setAllowedTypes(self::OPTION_RESPONSE_FALLBACK_ON_ERROR, ['boolean']);
```
campru, line 78: The type should be 'bool' instead of boolean. Symfony itself has defined in this way. Checking the internal code of the resolver, if I didn't miss anything, when checking the type does this
```
private static function isValueValidType($type, $value)
{
    return (\function_exists($isFunction = 'is_'.$type) && $isFunction($value)) || $value instanceof $type;
}
```

andreskrey: `boolean` it's an alias of `bool`: https://github.com/symfony/options-resolver/blob/3.4/OptionsResolver.php#L89

Aliases are available in the latest version of Symfony as well.

campru: I see they also have this for symfony 2.8 https://github.com/symfony/options-resolver/blob/2.8/OptionsResolver.php , so we are safe

campru:
The changes have been reviewed and are fine. Could you add here a architecture decision record for this?

andreskrey:
Done

Docs/architecture/decisions/002-deprecate-response-fallback.md
```diff
@@ -0,0 +1,32 @@
...
+We decided to deprecate the fallback **in the next major version**. 
+
+A new parameter was introduced in the Producer configuration (`response_format`) that will define the expected response type, and this type is passed to the serializer to instanciate the appropiate visitor. For example, a PlainTextVisitor could be implemented that can process non-JSON responses while still producing **a mappable/processable object**.    
+
+## Consequences
+
+In order to introduce this change in a backwards compatible way, a new boolean option was introduced in the Producer configuration (`response_fallback`) that will prevent the fallback. Setting it to `false` **will throw an exception** when the payload cannot be deserialized.
+
+## Metadata
+Authors: @andres.rey,
```
campru, line 30:
```suggestion
Authors: @andres.rey
```

Docs/architecture/decisions/002-deprecate-response-fallback.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Decision
+
+We decided to deprecate the fallback **in the next major version**. 
+
+A new parameter was introduced in the Producer configuration (`response_format`) that will define the expected response type, and this type is passed to the serializer to instanciate the appropiate visitor. For example, a PlainTextVisitor could be implemented that can process non-JSON responses while still producing **a mappable/processable object**.    
+
+## Consequences
+
+In order to introduce this change in a backwards compatible way, a new boolean option was introduced in the Producer configuration (`response_fallback`) that will prevent the fallback. Setting it to `false` **will throw an exception** when the payload cannot be deserialized.
```
campru, line 27:
```suggestion
In order to introduce this change in a backwards compatible way, a new boolean option was introduced in the `RestConfigurableProtocol` (`response_fallback`) that will prevent the fallback. Setting it to `false` **will throw an exception** when the payload cannot be deserialized.
```

Docs/architecture/decisions/002-deprecate-response-fallback.md
```diff
@@ -0,0 +1,32 @@
...
+
+Accepted
+
+## Context
+
+Currently the RestConfigurableProducer will pass the payload received from the target system to the serializer. If this one fails, it will catch the exception, fallback to the original response, **and continue the execution**.
+
+This was done to allow certain APIs to return a non-deserializable response while still being able to continue the execution, either **by ignoring the payload** (i.e.: when the response code is enough) or to map it to something else.
+
+This makes the mapper vulnerable, as it has no guarantee that what it will expect is something **that can be mapped**.
```
campru, line 15:
```suggestion
This makes the evaluator vulnerable, as it has no guarantee that what it will expect is something **that can be evaluated**.
```

Docs/architecture/decisions/002-deprecate-response-fallback.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Status
+
+Accepted
+
+## Context
+
+Currently the RestConfigurableProducer will pass the payload received from the target system to the serializer. If this one fails, it will catch the exception, fallback to the original response, **and continue the execution**.
+
+This was done to allow certain APIs to return a non-deserializable response while still being able to continue the execution, either **by ignoring the payload** (i.e.: when the response code is enough) or to map it to something else.
```
campru, line 13:
```suggestion
This was done to allow certain APIs to return a non-deserializable response while still being able to continue the execution, either **by ignoring the payload** (i.e.: when the response code is enough) or to evaluate it to something else.
```

Docs/architecture/decisions/002-deprecate-response-fallback.md
```diff
@@ -0,0 +1,32 @@
...
+
+This makes the mapper vulnerable, as it has no guarantee that what it will expect is something **that can be mapped**.
+
+Due to the fallback is present by default, one endpoint **could randomly fail** if the target system decides to send back a response in a format that is not expected by this bundle.
+
+## Decision
+
+We decided to deprecate the fallback **in the next major version**.
```
campru, line 21:
```suggestion
We decided to deprecate the response fallback in this version and remove it **in the next major version**. 
```

Docs/architecture/decisions/002-deprecate-response-fallback.md
```diff
@@ -0,0 +1,32 @@
...
+
+This makes the mapper vulnerable, as it has no guarantee that what it will expect is something **that can be mapped**.
+
+Due to the fallback is present by default, one endpoint **could randomly fail** if the target system decides to send back a response in a format that is not expected by this bundle.
+
+## Decision
+
+We decided to deprecate the fallback **in the next major version**. 
+
+A new parameter was introduced in the Producer configuration (`response_format`) that will define the expected response type, and this type is passed to the serializer to instanciate the appropiate visitor. For example, a PlainTextVisitor could be implemented that can process non-JSON responses while still producing **a mappable/processable object**.
```
campru, line 23:
```suggestion
A new parameter was introduced in the `RestConfigurableProducer` (`response_format`) that will define the expected response format, and this format is passed to the deserialize function to instantiate the appropriate visitor. For example, a PlainTextVisitor could be implemented to process non-JSON responses while still producing **a evaluable object**. By default, if `response_format` is not setup, `encoding` will be passed.
```

Docs/architecture/decisions/002-deprecate-response-fallback.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Context
+
+Currently the RestConfigurableProducer will pass the payload received from the target system to the serializer. If this one fails, it will catch the exception, fallback to the original response, **and continue the execution**.
+
+This was done to allow certain APIs to return a non-deserializable response while still being able to continue the execution, either **by ignoring the payload** (i.e.: when the response code is enough) or to map it to something else.
+
+This makes the mapper vulnerable, as it has no guarantee that what it will expect is something **that can be mapped**.
+
+Due to the fallback is present by default, one endpoint **could randomly fail** if the target system decides to send back a response in a format that is not expected by this bundle.
```
campru, line 17:
```suggestion
Due to the fallback is present by default, one endpoint **could randomly fail** if the target system decides to send back a response in a format that is not expected by the application.
```

andreskrey: Isn't framework a bundle? You cant run framework on its own. Why do you want to change it to application?

campru: Because it depends on how the producers are configured, and this is not configured in the framework itself

PR #129, opened 2020-05-05
Feature: Return message summary if it contains UTF8 characters

andreskrey:
I've noticed that if a request throws an exception and if the response body includes the character `’`, the function `get_message_body_summary()` **will return null**.

The expected result should be the string including **the special character**.

This happens due to the regex used to match any printable character lacks of the `/u` flag.

For example, this returns `1`, forcing `get_message_body_summary()` to return `null`:
```
preg_match('/[^\pL\pM\pN\pP\pS\pZ\n\r\t]/', '’')
```

Other multibyte characters like ௵ and ဪ also trigger this issue.

Adding the `/u` flag solves this issue.

### DIFF
--- a/Components/WebService/Rest/Middleware.php
+++ b/Components/WebService/Rest/Middleware.php
@@ -57,7 +57,7 @@ class Middleware
             $summary = mb_substr($summary, 0, self::$truncateResponseSize).' (truncated...)';
         }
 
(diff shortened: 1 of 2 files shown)


andreskrey:
Same merge request was submitted to https://github.com/guzzle/psr7/pull/330

PR #134, opened 2022-04-04
Feature/delay interceptor

jams246:


Core/Handlers/MessageHandler.php
```diff
@@ -405,6 +407,9 @@ public function onHandleException(
                 $this->dispatchEvent($exchangeBackup, HandlerEvent::THROTTLING_HANDLE_EVENT_NAME);
 
             } // If it's an exchange that can be retried later but it's failing due to an error
+            elseif ($originalException instanceof DelayException) {
+                throw new UnrecoverableRestException('Delay Interceptor triggered.');
```
andreymor, line 411: Should not this be an event triggered instead of an exception?
From what I got this exception will exist if delay interceptor needs to be executed, right?

jams246: This was part of a previous iteration.

It has been updated with the event dispatch.

https://github.com/smartboxgroup/integration-framework-bundle/pull/134/files#diff-373b6979369f09a295489b6d1459d3deab4813fedd108371d388988f5daa48a9R412-R416

andreymor (approved):

